PR #2, opened 2022-04-23
Update help files to the latest

mkilgore:
Not sure what happened but the ./internal/help files seem to be fairly
old compared to `development` in the old qb64 repository. This updates
them all to match that repository, let me know if this looks sane. I
made sure to keep the small changes already made to them in this
repository.


SteveMcNeill:
This isn't something we'll want to do.

What happened is this:  

QB64 pulls pages from the online wiki, and those become the Help files.  The old repo has help files from the old wiki site... which unfortunately, we don't have access to.   (RC put a version up on internet archive, but just like with the forum files he put up, it's missing important information so we can rebuild/import directly from it.)  Instead, we had to use the last copy of someone's off-line version of the wiki and use it as a reference to rebuild a new wiki.  

The wiki we built was based off a web crawl about 3 months ago, and that's what we're using as the base for the new wiki that we're working and expanding upon.  Anyone who clicks the "Update Help" pages in the Help Menu will come to the new site and grab the files which I'd uploaded previously for us into the repo.  These files are, indeed, older than the ones in the old Help index, but they're what we're building off from and moving forward with.  

It'd be best to leave these newer "old" files in the repo, rather than include the older "more recently edited" files, as they're the ones that actually match our current wiki.

What we need to do here, to truly address the issue, is to compare file vs file, and see where these older help files contain more/different information than the new ones, and then we can update the current wiki with whichever has the best contents for the pages in question.  At that point, a click on "Update Help" would resolve the issue and then we could update what comes in the repo as a default to begin with.

If all that makes sense.  LOL

SteveMcNeill:
We need to update the wiki itself, before we update these to something different than what the wiki contains.

PR #3, opened 2022-04-23
Add missing mingw files

mkilgore:
These are files missing from the mingw32 and mingw64 directories,
along with a few other files missing from internal/c. The files
themselves came from the top of `development` in the previous qb64
repository.

---
If you excluded these files because they're unnecessary definitely let me know :) Some of them like those in `bin` are definitely necessary though, and I'm guessing a largely unpredictable set of the `lib` files are necessary as well, so probably best to just keep it all for the time being. Eventually it would be good to get this out of the repository completely.


SteveMcNeill:
These look like all *.exe, *.o, *.a files which would've been blocked via the .gitignore.  They should definitely probably be in there -- though I'm quite curious just how the heck QB64 is compiling and working without any of them!  O-o!  Definitely feel free to pull these into the repo.

SteveMcNeill:
Merged in missing files.

PR #4, opened 2022-04-23
Remove LICENSE file at root of repository

mkilgore:
COPYING.txt and licenses already do a pretty good job of explaining the
situation. As not all the code in the repository is covered under MIT we
probably shouldn't just have another license file in the root of the
repository with no explanation.

---
This is mostly just me being picky :P But I think it's a bit misleading and we already have the clearer information sitting right besides it.


SteveMcNeill:
I agree here.  We have a license folder.  Don't need the file in the front of things to clutter stuff anymore than necessary.

SteveMcNeill:
Removed redundant license info.

PR #5, opened 2022-04-23
Mark ./setup_lnx.sh as executable

mkilgore:
git is capable of keeping track of the executable bit on files, adding
it to the setup script allows those who clone the repository from Linux
to run the script without having to add the bit themselves.

---
There might be others that need to be marked but I think this is the important one for now.


mkilgore:
I was wrong and there's more than a few other scripts that need to be marked executable to be able to build :-/ I'm going to fix them all in the automated build PR I'm working on rather than here. Just marking ./setup_lnx.sh as in this PR doesn't help anything because the script just tries to run other non-executable scripts and doesn't work.

PR #6, opened 2022-04-23
Add automated build

mkilgore:
This imports the basic logic from the previous build. All it does is
compile a version of QB64 using ./internal/source, and then uses that
QB64 as a 'bootstrap' step to compile `qb64.bas` directly and produce a
proper version of qb64.

After being compiled a briefly tested, teach version is packaged up and attached as a build artifact to the build, so every build (PRs and others) will have produced copies people can try. This should aide in making it easier to test PRs before merging.

Additionally, we can now easily create releases by simply creating a tag in the format of `vX.X.X`. That will trigger a build to happen which will build the current version denoted in `Version.bas` and create a draft release with the release version of QB64 attached to it. Along with that, we should be updated `Version.bas` when necessary to bump the version for the next release.

The version format is also a bit different now. We are using a SemVer compatible version, with `major.minor.patch` for the main version and then a tag in the format of `-X-hash` for all non-release builds, where `X` is the number of commits since the last release, and `hash` is the first 8 characters of the commit hash. That way makes it easy to identify where a development version someone is using came from.


SteveMcNeill:
I don't honestly know a thing about the automated build process which github uses.  I'd like to see it up and going, but don't know enough about the process to offer any decent insight on whether this is what we need or not.  You guys will have to use your best judgement here in this case.  When everything's ready to go with it, then go with it.  I trust you guys know more about this than I do.  :)

SpriggsySpriggs:
I, too, don't know anything about the automatic build process

mkilgore:
Not a problem :) I think I have a pretty good understanding of their previous process. I'm still working on it but when I have it functional I can explain how it works.

PR #8, opened 2022-04-25
Update Install for windows, replace compilers

SteveMcNeill:
Replaced the mingw compilers to the latest stable version available (both are smaller than the old versions now).

Compressed the mingw compilers to self-extracting EXE files for windows.

Changed windows set up batch file to run the self-extracting EXE.

Old folders are now gone, and we ship with the compilers already compressed, so multiple installations on someone's PC will take up considerably less disk space.


.gitignore
```diff
@@ -1,17 +0,0 @@
-internal/c/c_compiler
```
mkilgore, line 1: Please add this file back :)

mkilgore:
Two other thoughts:

1. I don't see `setup_win.bat` changes, I'm guessing you meant to include them.
2. I think we should put them somewhere other than `./internal/c/c_compiler`, like maybe a `./internal/c/mingw` folder. The idea being that it'll be annoying if deleting `./internal/c/c_compiler` also deletes the installers tracked by `git`.

SpriggsySpriggs:
Might be worth making a branch for testing and then merge changes from there as they get verified as working

PR #10, opened 2022-04-26
Add back default resource files, update readme,  ...

SteveMcNeill:
...correct wiki pointing back at old site

undoing part of the undo which I did with the last undo...  Gah!!

I hate github, and I think it feels the same about me!

Anyways, I think this corrects everything that was reverted out and shouldn't have been, as well as (once again) pointing help to the proper place.


SteveMcNeill:
I'm thinking this cleans up all the stray files and restores all the proper files to what they should be.

At this point, the help should once again work properly 
         -- NOTE:  using https://wiki.qb64forum.com as the wiki link won't work.  Some sort of 301 error is generated when forwarding from the subdomain, so the proper link of https:/www.qb64phoenix.com/qb64wiki 

Added the resource folder back in place, with a few base audio, image, font, word lists for users to use with examples and when viewing code from the wiki.

If there's anything else missing in here, I'm not seeing it at the moment.

SteveMcNeill:
Shelving this to avoid conflicts with Matt's PR.  Will see what's necessary to update after his goes through.  We were both addressing a bunch of the same issues. ;)

PR #11, opened 2022-04-26
Fix repository, and reintegrate mingw executables.

mkilgore:
This accomplishes two things:

1. Resets the entire repository to match 0cc86283
2. Re-adds Steve's changes from #8 and #9, along with my suggestions on the first PR.

This should get us back to a good working state, and also puts it in a place where I can add the automated build.


mkilgore:
I believe this is good to go now :)

SteveMcNeill:
Looking good then.   Here goes with pull #X to try and swap in the new compilers and setup.  :)

PR #13, opened 2022-04-28
Updated Wiki URL to qb64phoenix

mechatronic3000:
-Updated internal help Wiki address to https://qb64phoenix.com/qb64wiki. 
-You will need to either update or delete config.ini in the /internal folder.


mkilgore:
Hi! Thanks for the PR, but I believe this is actually already fixed. That file got moved at some point in the past and we already updated the wiki address [here](https://github.com/QB64-Phoenix-Edition/QB64pe/blob/main/source/global/IDEsettings.bas#L221) a few days ago.

mechatronic3000:
Hello

My apologies, apparently I failed to make sure I had the latest dev
version. I will try to do better next time.

Thanks

PR #14, opened 2022-04-28
Remove exes from packages and repository, redo distribution creation, add automated ./internal/source updating

mkilgore:
This ended up compiling a few different things:

1. Per @SteveMcNeill suggestion, the mingw*.exe's are removed from the repo. Now the `setup_win.bat` downloads the exe's and runs them. This happens during the build as well, so the portable Windows version of QB64 still contains the compiler inside.
2. The distribution logic was fixed to no longer be including the mingw*.exe's inside, which was bloating the size. The new logic makes it a lot easier to ensure no extra files get included.
3. I added some tests that verify the created distributions work as expected, by running the setup scripts and verify qb64 is properly compiled and itself can compile a program.
4. I added the automatic `./internal/source` updating. It will happen on every merge to `main` if compilation results in changes to `./internal/source`.


SteveMcNeill:
Looks to me like this should be what we were looking for.  The repo should be a lot smaller and a lot less cluttered after these changes go through.  :)

SteveMcNeill (approved):
Looks to me like this should be what we were looking for. The repo should be a lot smaller and a lot less cluttered after these changes go through. :)

PR #18, opened 2022-04-28
Add new tests to CI

mkilgore:
This adds two new kinds of tests to CI:

1. 'Compilation' tests - these tests require a variety of dependencies, and thus exercise that all of the dependencies required by QB64 work (QB64 itself only requires a small subset of the functionality QB64 has).
2. 'QBasic' tests - these are basically just a stripped down version of the 'samples' folder that used to exist. I've taken out most of the large files and am simply verifying that all of the `.bas` files correctly compiling using QB64.

Unfortunately it is a bit slow because there's over 100 QBasic tests, but they're definitely useful and I don't think they take long enough to be concerned about it. On my machine they take 10 minutes, so that adds up to about a 20 minutes runtime for our build. If we do want to shorten it we could probably prune some of the `qbasic_testscases` files without really losing much in the way of testing.


SteveMcNeill (approved):
Only 698 files to view...   No problem!

Looks like a good batch of basic tests to me!

PR #36, opened 2022-05-06
Generate icon.rc file on all platforms and copy icon file into temp directory

mkilgore:
This is some enabling work to make #33 (make-based build system) easier. The end goal is to simplify/unify how QB64 gets built so that the same Makefile that creates compiled programs can also build QB64 itself from the `./internal/source` folder. To do this I would like to be able to use `$EXEICON` and `$VERSIONINFO` to specify the Windows resource information in the QB64 source itself. This would remove the need to manually ad it after-the-fact as is currently done in `.ci/bootstrap.bat`.

Currently using those metacommands in `qb64.bas` is impossible for two reasons:

1. The Windows resource files are only created on Windows, but `./internal/source` is created on Linux and used for all platforms.
2. The generated `icon.rc` file from `$EXEICON` includes an absolute path to the icon file, which isn't portable when building `./internal/source` on a different system.

This PR fixes both of these problems. The first one is relatively simple, the code was simply changed so that the `icon.rc` file is generated on all platforms even if it won't be used. The second is also fairly simple, I wrote a new `CopyFile&()` function and simply copy the icon file directly into the temp directory. This allows the `icon.rc` file to use a relative path to the icon file since it is located in the same directory as it.

There are also some miscellaneous changes in here:

1. `./tests/run_tests.sh` was fixed to actually fail the build on test failures (oops...)
2. I moved some file-related utility functions from `qb64.bas` into `utilities/file.bas`
3. I added a few more tests for `$EXEICON` and `$VERSIONINFO`
4. I added a test that verifies the produced `qb64.exe` on Windows has Windows resource information (IE. Icon, file properties).


SteveMcNeill (approved):
Everything seems in place to me here.  (Not that I'm much of an expert when it comes to the building test process.  LOL!)

PR #37, opened 2022-05-06
Run `clang-format` on all C and C++ files

mkilgore:
`clang-format` is a pretty handy tool that can format C and C++ files automatically according to a given configuration file describing the format. Thus I added such a file (`.clang-format`) to our repository and then ran `clang-format` on all of the C/C++ files I consider us to 'own' (IE. I ran it on all the code except for the external libraries we use). I also reviewed the changes manually (quickly) and added a few comments to turn the formatting off on some specification locations where the existing formatting is easier to read (just a few large arrays and lists in `libqb.cpp`).

I made sure that the files were all clean before I formatted them, so this shouldn't change any of the actual code. The big thing to review is simply the style itself and let me know if you'd like to see anything different. The style I went with is just the one I prefer.


SteveMcNeill (approved):
All these need a nice formatting.  Clang certainly couldn't make them any messier than they already were!  ;)

PR #51, opened 2022-05-12
Replace existing build system with make

mkilgore:
This replaces all of the various existing scripts and compile lines with a single `Makefile`. This has a few big benefits:

1. Drastically reduces the complexity and duplication of the compilation lines. We have way too many files that include random `g++` lines that are slightly different from each other, and that along with ones in qb64.bas makes it very hard to handle any kind of build changes.
2. QB64 itself can be built using the exact same build scripts as regular programs, by simply passing `BUILD_QB64=y` to make.
3. Makes it extremely easy to start splitting up the C++ code, as `make` can be used to easily compile separate source files and link them all together in the end without us having to do much work.

Obviously this is a fairly large change. I add a fair amount more testing to verify this change does work - I have tests for every different optional dependency, along with `DATA`, `$EXEICON`, `$CONSOLE:ONLY` and some others. The biggest potential question mark is weird `DECLARE LIBRARY` usages that I wasn't able to make tests for (I may have gotten them all, but it's somewhat poorly documented so I'm not so sure.

---

For the moment I'm going to leave this PR as a draft and ask some on the forum to give this a test as a bit of a sanity check. Please give the PR a look when you have time :)

Closes: #33 

Edit: Note that due to `./internal/source` still housing a QB64 that uses all the existing build scripts, they can't be removed in this PR. It doesn't matter that much though as the `make` logic doesn't care if they are still there, and I'll submit a separate PR after this one to remove all those scripts after `./internal/source` is updated to use `make`.


internal/c/libqb/build.mk
```diff
@@ -0,0 +1,10 @@
+
+QB_LIBQB_SETUP_OBJS := $(PATH_INTERNAL_C)/libqb.o
+
```
mkilgore, line 3: I realized after looking it over that this file isn't actually getting used. I will be using this file in upcoming changes to work on splitting apart `libqb` but I'll remove it from this PR.

Makefile
```diff
@@ -0,0 +1,346 @@
...
+
+# The extra tag to put on ./internal/temp and qbx.o when multiple instances are involved
+# This is blank for the 'normal' files
+TEMP_ID ?=
+
+# Extra flags go at the beginning of the library list
+#
+# This is important for libraries, since they could potentially be referencing
+# things from our dependencies
+CXXFLAGS += $(CXXFLAGS_EXTRA)
```
mkilgore, line 17: Note that related to #53 the new Makefile approach does not strip the executables, it's missing that flag. I should add that back as not doing that adds a significant size to the executables, but I'm thinking I may just try to do #53 as part of this change.

SteveMcNeill (approved):
Nothing glaringly obvious stands out to me as being off here.  I think you're good to go, when you feel ready to pull and merge.  ;)

PR #58, opened 2022-05-15
Add error handling to setup_win.bat, fix objcopy command

mkilgore:
The `setup_win.bat` was not previously reporting an error if the compilation fails, so I changed it to indicate that something went wrong and report an exit code. This has the benefit that it will fail the build if there is something wrong with the `setup_win.bat` test.

Also, the `objcopy` is bugged on Windows and does not work on OSX, so I'm going to modify the logic to work on Windows and simply remove the functionality from OSX.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,7 +89,7 @@ ifdef BUILD_QB64
 	else
 		EXE ?= qb64
 	endif
-else
+endif
 
 ifneq ($(filter clean,$(MAKECMDGOALS)),)
 	# We have to define this for the Makefile to work,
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -84,7 +84,11 @@ else
 endif
 
 ifdef BUILD_QB64
-	EXE ?= qb64
+	ifeq ($(OS),win)
+		EXE ?= qb64.exe
+	else
+		EXE ?= qb64
+	endif
 else
```
mkilgore, line 92: Looking at the logic this `else` isn't actually a bug, but it's quite odd looking due to the indentation. I'm pretty sure we can change this to `endif` as I intended (and remove the `endif` on line 100) and the logic works just fine.

SteveMcNeill (approved):

PR #60, opened 2022-05-16
Fix to _desktopwidth, _desktopheight, _screenicon

SteveMcNeill:
Fixes the issue as brought up on the forums here: https://qb64phoenix.com/forum/showthread.php?tid=408

Also added a small set of logic so we don't end up inside an endless loop if the screen is hidden (via _SCREENHIDE), or if it doesn't exist for whatever reason.

### DIFF
--- a/internal/c/libqb.cpp
+++ b/internal/c/libqb.cpp
@@ -27356,7 +27356,7 @@ void sub__icon(int32 handle_icon, int32 handle_window_icon, int32 passed) {
 
 int32 func_screenwidth() {
 #ifdef QB64_WINDOWS
-    return GetSystemMetrics(0);
+    return GetSystemMetrics(SM_CXSCREEN);
 #else
 #    ifdef QB64_GLUT
     return glutGet(GLUT_SCREEN_WIDTH);
@@ -27368,7 +27368,7 @@ int32 func_screenwidth() {
 
 int32 func_screenheight() {
 #ifdef QB64_WINDOWS
-    return GetSystemMetrics(1);
+    return GetSystemMetrics(SM_CYSCREEN);
 #else
 #    ifdef QB64_GLUT
     return glutGet(GLUT_SCREEN_HEIGHT);
--- a/internal/c/libqb.cpp
+++ b/internal/c/libqb.cpp
@@ -27355,25 +27355,11 @@ void sub__icon(int32 handle_icon, int32 handle_window_icon, int32 passed) {
 #endif // DEPENDENCY_ICON
 
 int32 func_screenwidth() {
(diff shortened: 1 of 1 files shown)


internal/c/libqb.cpp
```diff
@@ -27356,13 +27356,19 @@ void sub__icon(int32 handle_icon, int32 handle_window_icon, int32 passed) {
 
 int32 func_screenwidth() {
 #ifdef QB64_GLUT
-    while (!window_exists) {
-        Sleep(100);
-    }
+    //while (!window_exists) {
```
mkilgore, line 27359: Don't leave commented code in the file :) If someone wants to see what was there before they can check the `git` history.

internal/c/libqb.cpp
```diff
@@ -27396,14 +27408,21 @@ int32 func_screenheight() {
...
-    while (!window_handle) {
+
+    while (!window_handle && waited_enough > 0) {
         Sleep(100);
+        waited_enough = waited_enough - 1; //same here.  Endless loops are a bad no-no!
     }
 #    endif
+    if (waited_enough == 0)
+        return; // if the screen isn't created and doesn't have a handle after two seconds, then just return and don't try to minimize the window.
     glutIconifyWindow();
```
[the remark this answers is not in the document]
mkilgore: I created #66 for this.

internal/c/libqb.cpp
```diff
@@ -27355,51 +27355,23 @@ void sub__icon(int32 handle_icon, int32 handle_window_icon, int32 passed) {
...
-        Sleep(100);
-        waited_enough = waited_enough - 1;
-    }
-    if (waited_enough == 0) return 0; //if the screen isn't created and doesn't have a handle after two seconds, then return a value of 0.  Don't run an endless loop!
-#    endif
-    return glutGet(GLUT_SCREEN_WIDTH);
-#else
-#    ifdef QB64_WINDOWS
+#ifdef QB64_WINDOWS
     return GetSystemMetrics(0);
```
mkilgore, line 27359: It's a small thing, but there are constants you're intended to use for `GetSystemMetrics()` rather than hard-coded values. You should try swapping `0` for `SM_CXSCREEN` and `1` for `SM_CYSCREEN`, it makes it much easier to know what's going on. Hopefully those constants are already available in this file via the `#include`s, if they can't be found then I'm fine leaving it as-is :)

mkilgore (approved):
I'd like you to give the `SM_CXSCREEN` and `SM_CYSCREEN` thing a try if you don't mind, but otherwise looks good to me :)

PR #67, opened 2022-05-17
Fix to endless loop in retrieving search History

SteveMcNeill:
Replaced both instances with one single sub to reduce code and make things easier to maintain and upkeep.  It's always easier to work in one spot than it is in multiple.


source/ide/ide_methods.bas
```diff
@@ -20013,6 +19983,19 @@ FUNCTION GetBytes$(__value$, numberOfBytes&)
...
 END FUNCTION
 
+SUB RetrieveSearchHistory (SearchHistory() as string)
+    fh = FREEFILE
+    OPEN ".\internal\temp\searched.bin" FOR BINARY AS #fh
+    redim _preserve SearchHistory(1 to 10000) 'large initial array to hold the data
+    Do until eof(fh)
+       ln = ln + 1
+       line input #fh, SearchHistory(ln)
+       if ln > ubound(SearchHistory) then redim _preserve SearhHistory(1 to ln + 10000) 'large resize, it necessary
```
mkilgore, line 19993: Should this be `>=`? Since `ubound()` is the last index, if `ln` is larger than that then I think there's a problem on a `Line Input` line.

mkilgore (approved):

Issue #63, opened 2022-05-16
A regular non-release build should include a version label indicating this


mkilgore:
Currently the way versioning works is that the version label is generated during the CI build and placed into a file. This means that all CI builds have proper version labels on them to easily identify what version is being used.

Unfortunately this means that if you just build the repository manually you do not get a version label, making it confusing what exactly you're using if you're looking to get help. We should change the setup a bit so that if QB64 does not detect it was a release build (probably based on the presence of some file in `./internal`) then it produces some kind of generic "-LocalBuild" version label to indicate this.

PR #68, opened 2022-05-18
Report if the version of QB64 being used is unknown/manually-built

mkilgore:
When building directly from the repo (either from a git clone or a
download of the zip of the repository) the version reported is very
misleading because it will not have a version label, suggesting it is
actually a 'release' version when in fact it could be anything.

The ./.ci/calculate-version.sh logic is already setup to delete an
existing ./internal/version.txt during a detected release build, so we
can just place one in the repository and it won't impact the versioning
of CI and release builds, but will show up when building locally.

I also included some cleanup of ./internal/source files we don't need.
The ./.ci/compile.sh was changed a bit to make sure they don't get
added back by the automated push.

Fixes: #63


mkilgore:
If I understand you correctly, then no, because "-UNKNOWN" is only for the event that the script that normally determines the version information is not run. Basically when someone downloads the repo itself and builds it, rather than a CI or Release build.

I think you might be able to see the problem if you just go check one of your random QB64's that you've built locally. The 'About' pages still says `0.6.0` even though it's definitely not the `0.6.0` release, it's a random build of `main` or some other branch. The issue is that this is confusing for users who mistakenly downloaded the repo thinking they're getting a release version, and instead now have some completely random version of QB64. Adding the "-UNKNOWN" on the end adds a clear indicator (to them, and us when helping them) that something is wrong and they didn't download a proper version of QB64.

SteveMcNeill (approved):

PR #72, opened 2022-05-18
Further fix to searched.bin issues

SteveMcNeill:
This should get rid of the 3rd endless loop from the searched.bin and github changing line endings.


internal/source/searched.bin
```diff
@@ -1,2 +0,0 @@
-
```
mkilgore, line 1: Just FYI, I also removed this file in #68.

mkilgore (approved):

PR #73, opened 2022-05-19
Clean up changes

SteveMcNeill:
Added to win setup to clean up the mingw compiler after installation.

Fixed the issue where some Linux distros won't clean up the exe when Run Only is selected in the menu bar.


source/qb64.bas
```diff
@@ -1086,10 +1086,10 @@ IF C = 9 THEN 'run
             IF os$ = "LNX" THEN
                 IF LEFT$(lastBinaryGenerated$, LEN(path.exe$)) = path.exe$ THEN
                     SHELL QuotedFilename$(lastBinaryGenerated$) + ModifyCOMMAND$
-                    SHELL _HIDE _DONTWAIT "del " + QuotedFilename$(lastBinaryGenerated$)
+                    KILL QuotedFilename$(lastBinaryGenerated$)
```
mkilgore, line 1089: Have you tested this? I probably can for you, but I don't think you need to add the quotes.

mkilgore (commented):
Looks fine, but unfortunately because you just reverted the SCase changes in a new commit it actually still screws up the history. A bit later I can help you fix it up :)

source/qb64.bas
SteveMcNeill: Aren't quotes used in Linux around a filename when passing it to KILL?    I've got to admit, I'm not usually a Linux programmer, so little syntax like this escapes me a lot of times.   I would think with KILL being a QB64 keyword, it'd handle things the same with Linux as it does with Windows, but I guess it might not??   I'll test it in a virtual linux and see how it behaves for certainty's sake.  ;)

mkilgore: I think it does work the same, I don't think quotes are required on windows either :P certainly QB45 doesn't require them when I tried it. Keep in mind this:

```
KILL "foobar.txt"
```

Doesn't have quotes in the string passed to kill.

Issue #64, opened 2022-05-16
Wiki Parsing Improvements


mkilgore:
Improvements to visual layout and various quirks.

PR #75, opened 2022-05-20
Wiki dev

RhoSigma-QB64:
Improved Wiki parser / built-in help
- ide_methods.bas, just improved file name validation, escaping invalid/critical chars, and added handling for external wiki links (opened in standard app)
- wiki_global.bas, some new SHARED variables and a list of UTF-8 exchange sequences
- wiki_methods.bas, practically turned every stone, its probably hard to compare against the old file version
- internal/help, updated files to contain latest wiki edits and to follow new naming scheme as descibed here https://discord.com/channels/975381912350752819/975389657212805160/976808241197625354

What' new
- wiki text is almost handled the same way as a browser would do, ie. code/text blocks are solid constructs, which reset styling in the end, hence forgotten closing text style tokens will no longer cause leaking of the current style past the block and into the regular text anymore
- blocks are separated from regular text with rulers, {{small|Code by User}} template text will be embedded in the bottom ruler, if noted right after {{CodeEnd}} or next line
- pages got a header with last updated date or other relevant messages
- \<center> and \<p style="align:center"... tags are no longer ignored, hence what's centered on the page viewed in the browser is now centered in the built-in help too
- external links are clickable now (open in standard app)
- more reliable UTF-8 char substitution, 2-byte, 3-byte and 4-byte sequences are properly detected now, missing chars can be added easily
- and more

Does close issue #64


SteveMcNeill:
A few glitches as I mentioned in Discord:
1) If links.bin is erased, we get an internal IDE error with F1.
2) There appears to be an ever-increasing list of files to download, if you click "Update All" multiple times.
3) Clicks up top on the tab area in help, where the white highlighted topics are listed, count as clicks inside the content area as well.
4) Same topic can be listed multiple times in the tab/index area of help.  For example "Help - Let - Help" has help listed twice.

RhoSigma-QB64:
Problems 1-3 are fixed now.

Problem 4:
- It did always work that way since the QB64 SDL age, nobody complaint about.
- It is obviously supposed to be a linear link history, which you can backstep with the backspace key, and not mainly thought to be a tab bar with once visited pages.
  - Both will use the same backtrace array internally, so stripping out double entries in the tab bar would break the linear history for the backspace key.
- Somehow I'll address this problem sooner or later, but certainly not in this PR anymore.

That's all there is to it!

source/ide/ide_methods.bas
```diff
@@ -19801,9 +19828,19 @@ FUNCTION findHelpTopic$(topic$, lnks, firstOnly AS _BYTE)
...
+        Help_IgnoreCache = 1
+        a$ = Wiki$("Keyword Reference - Alphabetical")
+        Help_IgnoreCache = 0
+        Help_ww = 78: WikiParse a$ 'assume standard IDE width for parsing
+    END IF
+    '----------
     a2$ = UCASE$(topic$)
     fh = FREEFILE
-    OPEN "internal\help\links.bin" FOR BINARY AS #fh
+    OPEN "internal\help\links.bin" FOR INPUT AS #fh
```
mkilgore, line 19843: Is there a reason `BINARY` didn't work? From my understanding it should be faster and functionally work the same in this situation.

source/ide/wiki/wiki_methods.bas
```diff
@@ -6,27 +6,29 @@ FUNCTION Back2BackName$ (a$)
...
+        SELECT CASE c
+            CASE 32 '                                    '(space)
+                PageName2$ = PageName2$ + "_"
+            CASE 34, 38, 42, 47, 58, 60, 62, 63, 92, 124 '("&*/:<>?\|)
+                PageName2$ = PageName2$ + "%" + HEX$(c)
+            CASE ELSE
+                PageName2$ = PageName2$ + CHR$(c)
+        END SELECT
+    NEXT
+    PageName3$ = wikiSafeName$(PageName2$) 'case independent name
```
mkilgore, line 25: Rather than name these `PageName$`, `PageName2$` and `PageName3$`, could we go with something like  `PageNameRaw$`, `PageNameExpanded$` and `PageName$`? With the idea being that the rest of the code can just use `PageName$` and be independent of however many steps we do here. If we also put the previous logic into a function then this could just be:

`PageName$ = wikiSafeName$(urlExpandSpecialCharacters$(PageNameRaw$)))`

source/ide/wiki/wiki_methods.bas
```diff
@@ -6,27 +6,29 @@ FUNCTION Back2BackName$ (a$)
...
-        i = INSTR(PageName2$, "&")
-        PageName2$ = LEFT$(PageName2$, i - 1) + "%26" + RIGHT$(PageName2$, LEN(PageName2$) - i)
-    LOOP
-    DO WHILE INSTR(PageName2$, "/")
-        i = INSTR(PageName2$, "/")
-        PageName2$ = LEFT$(PageName2$, i - 1) + "%2F" + RIGHT$(PageName2$, LEN(PageName2$) - i)
-    LOOP
+    'Escape all invalid and other critical chars in filenames
+    PageName2$ = ""
+    FOR i = 1 TO LEN(PageName$)
```
mkilgore, line 14: This seems like something that could be thrown in a `UrlExpandSpecialCharacters$()` function, or something along those lines :)

mkilgore (approved):
Overall this looks good to me, I had one very minor suggestion but I like the general improvements you made and also very much appreciate the through description :)

Issue #104, opened 2022-06-08
The build artifacts should have the version in their name


mkilgore:
I avoided this because it's a bit cumbersome to do, but currently the build/release artifacts are simply called `qb64_win-x64` or `qb64_lnx`, which no version in the name. This can get a little annoying because it means if you download these files and have it sitting around it's not obvious what version it is.

We should fix this by putting the version in the name, such as `qb64_win-x64-v0.8.0` or something along those lines. To do this we'll need to read the version out of the `./source/global/version.bas` file and also check `./internal/version.txt` if it exists.

PR #109, opened 2022-06-10
Add version to artifact name

mkilgore:
The artifacts like `qb64-lnx.tar.gz` will now also include the version
in their name, making them `qb64-lnx-X.Y.Z.tar.gz` instead.

Fixes: #104


RhoSigma-QB64 (approved):
It's very welcome change.

PR #114, opened 2022-06-12
Add separate libqb folder with new APIs, along with C++ automated testing.

mkilgore:
This is a reasonably big one but I didn't see a better way to split it up, this is the next step after #102. If you have any questions please ask :)

The `./internal/c/libqb/` folder now has two new folders `./internal/c/libqb/src/` and `./internal/c/libqb/include` for headers and C++ source files contributing to libqb. These source files are all compiled independently into `.o` files, and then linked into the final application. Which of these files gets built and linked is determined by the logic in `./internal/c/libqb/build.mk`.

The `threading-*.cpp` files show the general approach for platform-specific implementations.  Depending on exactly how different the implementations for Windows/POSIX/x64/x86/etc... are we can avoid all the `#ifdef` logic everywhere by simply providing completely different source files and building the necessary ones for that platform. This can work for dependencies too, so Ex. We can check if `DEP_SOCKETS` is true and the compile/link the necessary source files if that's the case.

---

The C++ automated testing is fairly simple and something I wrote a while back for a different project. All the logic is just a single file `test.cpp`, which takes in a list of functions and calls them one at a time, logging information as it goes. There's various `test_assert*` functions that check if a condition is true, and if not mark that test as failed and output some information on why it failed. The script `./run_c_tests.sh` can run all the C++ tests in a row and checks if any of them report failures. There's various improvements that could probably be made, but I think it's a good starting point for us.

The building of the C++ tests is a little complicated and located in `./tests/build.mk`, but think once I get a few more examples in there it will start to make sense. Ignoring the `make` magic, you basically just define the name of a test suite, and then what source files should be used to build that test suite. You can also provide any compiler flags for that test, if they are needed.

---

A notable detail that might get lost is that `CONDITION_VARIABLE` (which I need for the HTTP/HTTPS stuff that's coming up) on Windows is Vista and up, so while I think we were already not XP compatible this makes us more-so incompatible.

### DIFF
--- a/internal/c/libqb/src/threading-posix.cpp
+++ b/internal/c/libqb/src/threading-posix.cpp
@@ -1,0 +1,103 @@
+
+#include <stdlib.h>
+#include <string.h>
+#include <pthread.h>
+
+#include "mutex.h"
+
+struct libqb_thread {
+    pthread_t thread;
+};
+
+struct libqb_mutex {
+    pthread_mutex_t mtx;
+};
+
+struct libqb_condvar {
+    pthread_cond_t var;
+};
+
+struct libqb_mutex *libqb_mutex_new() {
+    struct libqb_mutex *m = (struct libqb_mutex *)malloc(sizeof(*m));
+    pthread_mutex_init(&m->mtx, NULL);
+    return m;
+}
+
+void libqb_mutex_free(struct libqb_mutex *mutex) {
+    pthread_mutex_destroy(&mutex->mtx);
+    free(mutex);
+}
+
+void libqb_mutex_lock(struct libqb_mutex *m) {
+    if (m == NULL)
+        return;
+
+    pthread_mutex_lock(&m->mtx);
+}
+
+void libqb_mutex_unlock(struct libqb_mutex *m) {
+    if (m == NULL)
+        return;
+
+    pthread_mutex_unlock(&m->mtx);
+}
+
+struct libqb_condvar *libqb_condvar_new() {
+    struct libqb_condvar *c = (struct libqb_condvar *)malloc(sizeof(*c));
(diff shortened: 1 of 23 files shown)


RhoSigma-QB64 (approved):
Uhh, that took quiet a bit of time. I really like your C++ programming style, clean and good readable. I've to admit I'm not that bright about this entire `Make` stuff, but from what I've seen here and from earlier PRs, I'd pretend you have mastered your craft. Approved.

internal/c/libqb/src/threading-posix.cpp
```diff
@@ -0,0 +1,105 @@
...
+    return m;
+}
+
+void libqb_mutex_free(struct libqb_mutex *mutex) {
+    pthread_mutex_destroy(&mutex->mtx);
+    free(mutex);
+}
+
+void libqb_mutex_lock(struct libqb_mutex *m) {
+    if (m == NULL)
```
mkilgore, line 34: This shouldn't be here, the Windows version doesn't have it and I don't think we should allow it to work this way anyway (silently not locking if the mutex is NULL)

internal/c/libqb/src/threading.cpp
```diff
@@ -0,0 +1,33 @@
...
+    comp->var = libqb_condvar_new();
+}
+
+void completion_clear(struct completion *comp) {
+    libqb_mutex_free(comp->mutex);
+    libqb_condvar_free(comp->var);
+}
+
+void completion_wait(struct completion *comp) {
+    libqb_mutex_lock(comp->mutex);
```
mkilgore, line 20: This should use the mutex guard, rather than manual lock/unlock.

RhoSigma-QB64 (approved):

PR #139, opened 2022-08-24
Avoid excessive disk access while typing

RhoSigma-QB64:
- No longer constantly (over)writes files in the `internal\temp` folder while typing in the IDE, as the generated C/C++ code is buffered internally now.
- Buffers are automatically written out to disk on a `Make` request (F5/F11).


SteveMcNeill (approved):
Everything appears to be functional here to me.  I don't see any obvious issues with this.  If something was missed, we'll patch it if needed later.  ;)

PR #133, opened 2022-08-09
miniaudio audio backend support

a740g:
Hi @mkilgore,

I am opening this PR like we discussed on Discord. Please let me know if you need my help with anything.

Issue #115, opened 2022-06-14
Add MIDI support to _SNDOPEN()


hackwrench:
No sounds currently mix and play one after each other. Sound effects/speech can't play in music.

mkilgore:
Can you give an example of what you're doing that doesn't work? I just did a sanity check with two `_SndPlayCopy`s and they play at the same time and get mixed as expected. Whether you get the mixing will depend on what commands you're using to play the sounds though.

hackwrench:
QB64pe still compiles the two parameter _SNDOPEN no longer apparently mentioned in the wiki.

mkilgore:
I see, I'm not sure why the wiki doesn't mention that parameter at all but I believe those settings are unnecessary now. A change was made to the backend of QB64 and all the settings are applicable to all calls to `_SNDOPEN()` (AFAIK anyway...).

Are you seeing a behavior difference due to this compared to previous versions, and can you give a more complete example that doesn't work?

I would note that for each sound you want to play mixed you need a new _snd handle, either via `_SNDOPEN()` or `_SNDCOPY()`.

hackwrench:
This file compiles, runs and doesn't play the MIDI file.
`Filename$ = "TONOWH~2.MID"
Select Case UCase$(Right$(Filename$, 4))
    Case ".WAV": SoundSet$ = "VOL,SYNC,LEN,PAUSE"
    Case ".OGG": SoundSet$ = "VOL,SYNC,LEN,PAUSE"
    Case ".AIF": SoundSet$ = "VOL,SYNC,LEN,PAUSE"
    Case ".RIF": SoundSet$ = "VOL,SYNC,LEN,PAUSE"
    Case ".VOC": SoundSet$ = "VOL,SYNC,LEN,PAUSE"
    Case ".MID": SoundSet$ = "VOL"
    Case ".MOD": SoundSet$ = "VOL,PAUSE"
    Case ".MP3": SoundSet$ = "VOL,PAUSE,SETPOS"
End Select
If SoundSet$ <> "" Then LoadSound& = _SndOpen(Filename$, SoundSet$)

MySound& = _SndCopy(LoadSound&)
_SndPlay MySound&
Do: Loop Until Not _SndPlaying(MySound&)
_SndClose MySound&
_SndClose LoadSound&`

hackwrench:
`Print "Press any key to start"
Do: Loop Until InKey$ <> ""
[...]

PR #143, opened 2022-08-27
a740g's miniaudio audio backend

mkilgore:
Like the title says this is @a740g's new miniaudio-based audio backend. The big advantage it provides is being MIT-based, removing the LGPL dependency from QB64-PE programs. There are some known potential issues surrounding `_SNDRAW` (mentioned on the previous PR #133) and potentially `_MEMSOUND()` support, they need GitHub issues made but I don't think they're significant enough details that they need to be addressed now. Additionally we may uncover more as it gets used :)

With that in mind I changed the build process slightly and made it possible to switch back to the old audio backend if necessary. This can be done via a new checkbox in the `Compiler Settings` menu (renamed from `C++ Compiler Settings`). Note that the setting defaults to using the **new** backend. This way, users can stay on the latest version of QB64-PE (which has many improvements unrelated to audio) even if they need to use the old audio backend for some reason. The checkbox can be seen below:

image

There are associated documentation updates with this that are in the works :) Presumably unless something comes up this will go into `v3.1.0`.


mkilgore:
I had forgotten what @a740g mentioned regarding MIDI support, it's quite cool but we may want to remove it from this PR and integrate it later. The big thing is that it's non-optional, so using the miniaudio backend carries about a ~2MB file size increase (testing on Linux). A bit more than half of that is the bundled soundfont. Additionally the soundfont filename is hardcoded in the event you want to supply your own.

The actual code for MIDI support works great, we just need to integrate it with my language changes in #115 that make MIDI support optional and provide a way to specify a soundfont at compile time. I'll do that in a separate PR.

Of that ~2MB, about .7MB of it is from everything else not MIDI. It seems like the addition of libxmp (MOD, S3M, XM, IT) is an extra .2MB, and then miniaudio itself accounts for the other .5MB. Within miniaudio, it seems like FLAC support is .2MB (and can be turned off). These are rough numbers, if wanted I could get more exact info on this.

On Windows the results are a bit less significant, with exe's being about ~1.7MB larger. That means that everything except MIDI contributes an extra .4MB compared to OpenAL, and stripping out the extra file type support probably makes this change file-size neutral (This happens because OpenAL is statically linked on Windows, but dynamically linked on Linux).

We should probably consider what file size increase is acceptable and then go from there, that would inform of us what we would need to consider making optional. If we do choose to make some of these formats optional I think we could probably have a metacommand to specify the sound file types to include support for, that could work out nicely and that way you only pay for what you need. For compatibility with existing QB64 we only need to continue supporting Wav, Mp3, and ogg, so we could leave those as the default.

a740g:
This is true. The way it is right now, the file size will only increase if any audio functions are used. But that also means that using a simple Beep will add all the extra bloat. I had thought about this and wrote the whole extra plugins thing in such a way that it could be easily `#ifdef`d out. 

So, just conditionally compiling the below line from `miniaudio_impl.cpp` can give us a whole lot of control of what gets linked into the executable.
```C
ma_decoding_backend_vtable *maCustomBackendVTables[] = {
    &ma_decoding_backend_vtable_radv2,
    &ma_decoding_backend_vtable_tsf,
    &ma_decoding_backend_vtable_modplay
};
```

We can also wrap the `#include`s just above this line to reduce compilation time.

I actually paid attention to your QB64-PE language extension in https://github.com/QB64-Phoenix-Edition/QB64pe/issues/115 😉. So, I am positive that we can easily wrap this new format plugins thing using the metacommand mechanism you are working on. Honestly, I think that is the way to go.

RhoSigma-QB64 (approved):

mkilgore:
I do like how you organized it such that it's easy to turn things off :) After talking about it a bit it doesn't seem like including everything except for MIDI is a huge deal, the file size increase isn't crazy large all things considered. So with that said I'll take MIDI support out of this PR but otherwise leave it as-is :)

a740g:
> Edit: We also need to update the `licenses` folder. I think I can handle that.

I actually had a few this collected. Sharing it here to get you started. :)

[license_libxmp-lite.txt](https://github.com/QB64-Phoenix-Edition/QB64pe/files/9438192/license_libxmp-lite.txt)
[license_miniaudio.txt](https://github.com/QB64-Phoenix-Edition/QB64pe/files/9438193/license_miniaudio.txt)
[license_stb_image.txt](https://github.com/QB64-Phoenix-Edition/QB64pe/files/9438194/license_stb_image.txt)
[license_stb_vorbis.txt](https://github.com/QB64-Phoenix-Edition/QB64pe/files/9438195/license_stb_vorbis.txt)
[license_tinysoundfont.txt](https://github.com/QB64-Phoenix-Edition/QB64pe/files/9438196/license_tinysoundfont.txt)

a740g (approved):

PR #150, opened 2022-08-29
Fix High-DPI awareness on Windows

mkilgore:
We started defining `_WIN32_WINNT` a little while ago to express that we
require Windows Vista or above for support. This enables us to access
some Windows Vista-only APIs. The fact that `WINVER` also needs to be
defined was missed, and it seems that defining one means the other no
longer gets defined automatically as it did before. Thus we're simplying
now also defining `WINVER` the same as `_WIN32_WINNT`.

This fixes High-DPI awareness and a few other things that were gated
behind WINVER checks.


SteveMcNeill (approved):
Tested and solves the issue with DPI Awareness on my machine.

PR #153, opened 2022-09-01
Add _ROR & _ROL support. Fix issues with _SHR & _SHL

a740g:
@SteveMcNeill @mkilgore @RhoSigma-QB64

Opening a PR for `_ROL` and `_ROL` functions for QBPE-64.
These are implemented using multiple C functions on the `LibQB` side. However, they are mapped to `_ROL` and `_ROL` on the BASIC side.

@mkilgore You had mentioned an issue with `_SHL` where shifting bits left beyond say a byte would spill those to the word side.

I've implemented `_SHR` & `_SHL` just like `_ROL` and `_ROL` now and the issue above is now fixed.

Update: All `_SHL` & `_SHR` specific changes have been removed from this PR as some critical things will need to be considered and behaviors defined.


mkilgore (commented):
IMO it would be good to add a bunch of test cases of the expected results of these functions, it would give us a good idea of how they behave (especially with type conversions) and also highlight any potentially incompatibilities with the existing `_SHR` and `_SHL`. These should be pretty easy to do, just create a `./tests/compile_tests/shifts` folder and inside create a `test.bas` and `test.output`. In `test.bas` you can do a bunch of tests like this:

```vb
$CONSOLE:ONLY

Dim b as _Byte
DIm i as _Unsigned Integer
b = -2
i = &HAA00

PRINT _SHR(b, 2); _SHR(b, 8); _SHR(b, 20); _SHR(b, 70); _SHR(b, -1)

FOR k = -100 to 100
    PRINT _SHL(i, k); _SHR(i, k)
NEXT k

' More tests, you get the idea
```

And then test.output should just contain the expected console output of the program:
```
10 20 30 40 50
1 2
2 3
3 4
...
```

Obviously those numbers aren't right, but I hope it makes enough sense :) I'd be happy to help you get stuff in place if you want. For checking it locally you can just compile your test program and run it. Additionally you could make another `./tests/compile_tests/rotations` for testing `_ROL` and `_ROR`.

_Ideally_ the `_SHR` and `_SHL` tests produce the same output on the current QB64 as they do on the fixed version, but I suspect that won't be the case. The tests would help us understand the differences.

RhoSigma-QB64 (changes requested):
In `subs_functions.bas` the names assigned to `id.n = ` should reflect the correct "CaMeL" case, as these are used to determine the correct casing if "Camel Case" is set in IDE>Options>CodeLayout,
hence `_RoL` and `_RoR` or `_Rol` and `_Ror`.

tests/compile_tests/rotations/test.output
```diff
@@ -0,0 +1,20 @@
+ 240 :11110000
```
mkilgore, line 1: I really like the output formatting :) Makes it pretty easy to verify by eye.

I think it would be good to add a few more test cases for signed integer variables, and also rotations that are negative and rotations that rotate past the size of the integer (Ex. `_ROR(a, -1)`, `_ROR(a, 8)`, `_ROR(a, 17)`).

mkilgore (approved):

RhoSigma-QB64 (approved):

Issue #66, opened 2022-05-17
`glutIconifyWindow()`, `glutHideWindow()`, `glutShowWindow()` should not be called outside of the GLUT thread


mkilgore:
It's not safe to simply call `glutIconifyWindow()`, `glutHideWindow()`, `glutShowWindow()` with no threading protection while we have another thread that handles the GLUT stuff. There was some half-hearted attempts by using busy waits on a global variable, but that's really not the way to handle threading problems like these.

In this case these functions are all quite simple because they do not give any values back, so likely the easiest solution is to simply have the existing QB64 functions (`sub__screenicon`, etc.) set some kind of flag that the GLUT thread can see and perform the appropriate action. I believe the `glutTimerFunc` would be the place to do this, since it gets called at regular intervals.

PR #160, opened 2022-09-04
Add WindowWidth and WindowHeight

SteveMcNeill:
Make a few C variables available in QB64 which tells us the width and height of our window.  This is useful for when the window is scaled, in case the user wants to calculate screen clicks/print, or position the window in relation to the rest of their screen.


internal/c/libqb.cpp
```diff
@@ -37148,6 +37148,26 @@ int32 func__resize() {
...
+        if (tempx < 320) return; // minimal window width of 320.
+        if (tempy < 240) return; // minimal window height of 240
+        
+        ScreenResize = 1;
+        ScreenResizeScale = amount;
+
+        resize_snapback = 0;
+        resize_pending = 1; // this says we're going to do a resize
+
+        glutReshapeWindow(tempx, tempy);
```
mkilgore, line 37166: I'm pretty sure we're not supposed to do this outside of the thread handling the GLUT logic. Will the window not get resized by the existing logic if you just wait for the next screen refresh after setting the various settings?

internal/c/libqb.cpp
```diff
@@ -37126,9 +37126,9 @@ qbs *func__droppedfile(int32 fileIndex, int32 passed) {
 void sub__resize(int32 on_off, int32 stretch_smooth) {
 
     if (on_off == 1)
-        resize_snapback = 0;
+        resize_snapback = 0; //ON
```
mkilgore, line 37129: What does `OFF` and `ON` mean in this context?

Also, FWIW, we should look to replace a lot of these with `enum`s, then we can just name the values it can be.

source/subs_functions/subs_functions.bas
```diff
@@ -3867,3 +3867,31 @@ id.arg = MKL$(UINTEGER64TYPE - ISPOINTER) + MKL$(LONGTYPE - ISPOINTER)
...
+id.hr_syntax = "_WINDOWHEIGHT"
+regid
+
+clearid
+id.n = qb64prefix$ + "WindowScale"
+id.subfunc = 2
+id.callname = "sub__windowscale"
+id.args = 1
+id.arg = MKL$(LONGTYPE - ISPOINTER)
+id.hr_syntax = "_WINDOWSCALE(amount)"
```
mkilgore, line 3895: Since this is a command, should it be `_WINDOWSCALE amount`? I'm basing that off of the other commands in this file :D

internal/c/libqb.cpp
```diff
@@ -37148,6 +37148,26 @@ int32 func__resize() {
     return 0;
 }
 
+void sub__windowscale(int32 amount) {
+    int32 tempx, tempy;
+        if (amount < 50) return; // don't scale the screen to less than 50% of it's original size!!
```
mkilgore, line 37153: Is there a reason to restrict the size?

And if we're going to do that, I think it would be good to clip the values to the lowest allowed rather than doing nothing. We could do the same with the `320x240` limit.

internal/c/libqb.cpp
mkilgore: Note that if you remove the GLUT logic that will also fix the compilation issues, since I assume the variables are available even in `$CONSOLE:ONLY` mode (though they wouldn't do anything). Maybe trying to use the commands should error in `$CONSOLE:ONLY` mode, but none of the other functions look like they do that so 🤷

internal/c/libqb.cpp
```diff
@@ -37148,6 +37148,26 @@ int32 func__resize() {
...
+    int32 tempx, tempy;
+        if (amount < 50) return; // don't scale the screen to less than 50% of it's original size!!
+
+        tempx = environment_2d__screen_width * amount / 100;  // basically our width of the screen
+        tempy = environment_2d__screen_height * amount / 100; // and this is the height of the screen
+        if (tempx < 320) return; // minimal window width of 320.
+        if (tempy < 240) return; // minimal window height of 240
+        
+        ScreenResize = 1;
+        ScreenResizeScale = amount;
```
mkilgore, line 37161: I'm not sure how `ScreenResizeScale` works, you set it to the `amount` to scale, but the only usage of that variable is `resize_auto = ScreenResizeScale;`, and `resize_auto` is just supposed to be either `0`, `1` or `2` to indicate stretch or smooth scaling.

Looking at it closer, I think the thing missing is how we tell the GLUT thread to resize to a set window size. It's hard to work out how to do it with only the existing logic, but it seems like it wouldn't be too hard to add something around where `resize_auto` is handled that just forces the window to a size we supply from here, and then the rest of the logic after that should just work out.

internal/c/libqb.cpp
```diff
@@ -37148,6 +37148,26 @@ int32 func__resize() {
     return 0;
 }
 
+void sub__windowscale(int32 amount) {
```
mkilgore, line 37151: How is this intended to play with the `_Resize` option? Since the scaling respects the aspect ratio I think we could keep the current setting in place of either `_Smooth`, `_Stretch`, or off. Setting the `ScreenResizeScale` value appears to wipe it out.

Also, should the scaling be maintained if a new `SCREEN` is put in place? I don't really know, but it seems like a problem if it doesn't because if you set the screen to a new image to change the size, it will resize the window to the unscaled size until you do another `_WINDOWSCALE` call. That resizing back to the small window might be visible to the user.

internal/c/libqb.cpp
SteveMcNeill: We're letting GLUT handle the resizing.  That's what the resize_pending = 1 is for.  At least, that's how I took it, when looking at the functions for RESIZE.  

As for the compilation issues, I wonder if there's a flag somewhere for the "Parts System" that I haven't found yet.   I'll dig into that deeper, but this should probably be a non-console command as it resizes the window and not the terminal.

internal/c/libqb.cpp
SteveMcNeill: It's counterintuitive in this case.  When we set $RESIZE:ON in a program (or _RESIZE ON), snapback is 0.  $RESIZE:OFF sets it to 1.

snapback is the process where our window snaps back to its original size, if we try to scale it with $RESIZE ON and then _RESIZE OFF.  We don't refuse to scale, which would make sense to me...  We just scale then reset/snapback.

source/subs_functions/subs_functions.bas
SteveMcNeill: I'll double check the syntax here.  You're probably right.  ;)

internal/c/libqb.cpp
SteveMcNeill: You have to restrict to some point.  A window can't be scaled to 0% of its size -- it wouldn't exist!  Or a negative percent...  My screen is -320 pixels by -240 pixels.  😱

There comes a point where we seg fault and crash, and it's hard to determine.  You have to keep room for your titlebar's minimize, maximize, close buttons and such.   We could go sans limit, and put a warning there -- user beware -- but I figure anything scaled down beyond 50% probably isn't readable or usable anyway.  

FWIW, _RESIZE sets a 640x480 snapback limit.  Below that, it resets the screen to that size after trying to downscale it.

internal/c/libqb.cpp
SteveMcNeill: This is a magically command/variable.  It must work somewhere in the core files, because it handles the resizing perfectly, whereas just setting resize_auto itself doesn't.  With resize_auto, you have to calculate those new window limits and store them yourself.  With ScreenResizeScale, it's handled automatically somewhere.

internal/c/libqb.cpp
mkilgore: Right, GLUT does the resizing, the question is where in the code we should call `glutReshapeWindow()`. Similar to #66, it's not safe to call it from outside of the rendering logic and it may randomly have problems. It's probably not to hard to fix here, we just have to tell the rendering logic that currently does the resizing (in `GLUT_DISPLAY_REQUEST()`) what to do and have it call `glutReshapeWindow()`, rather than doing it here. There's already some `glutReshapeWindow()` calls in there that do the snapback logic and such, it would be the same but resizing the window to our scaling instead.

In regards to `ScreenResizeScale`, I ran a search over all the code but all it does is set `auto_resize` equal to it. I'm pretty sure the magic part is the `glutReshapeWindow()` call, because that triggers the `GLUT_RESHAPE_FUNC` which changes a bunch of the variables like `display_x` and friends to the new size.

Spriggsy95:
How does this jive with the DPI awareness that I added in to QB64 in the past?

SpriggsySpriggs:
How does this jive with the DPI awareness stuff I added to QB64 previously?

SteveMcNeill:
I started on this before my surgery.  Got sidetracked healing...  I'm going to have to take a bit to go back and look over all this once again and renew my brain on what all was going on here.  I'm going to close this pull request for the moment, and I'll reopen it after delving back into this issue once again in the future (if no one else looks into it first and sorts it out for us).  ;)

PR #161, opened 2022-09-04
Fix _MOUSEMOVE when window is resized

mkilgore:
`sub__mousemove` is trying to use `x_scale`, `y_scale`, `x_offset`, and
`y_offset` to calculate where the mouse should be in the event the
window coordinates are different from the screen coordinates.

Unfortunately, all four of those variables are actually never set in the
program. The real scale values and offsets (in the event of
letterboxing) are stored in `environment_2d__` values. This change
switches `sub__mousemove` to simply use the correct values when
calculating the mouse position.

Because `x_scale` and `y_scale` are not used anywhere else I just
removed them completely. I wanted to remove `x_offset` and `y_offset` as
well but there are a few spots that make use of it. It must be a bug,
since they are never assigned values other than zero, but I'm not sure
if the correct fix for the other locations is to use the
`environment_2d__` value or do nothing, so I'm leaving them for now and
we can address them later.

I tested the fix with `_FULLSCREEN`, `_FULLSCREEN _STRETCH`,
`$RESIZE:ON`, `$RESIZE:STRETCH`, `$RESIZE:SMOOTH`, and
`_FULLSCREEN _SQUAREPIXELS`, and in all cases `_MOUSEMOVE` sends
the mouse to the correct location for the current `SCREEN`.

### DIFF
--- a/internal/c/libqb.cpp
+++ b/internal/c/libqb.cpp
@@ -6605,7 +6605,6 @@ int32 selectfont(int32 f, img_struct *im) {
 int32 nmodes = 0;
 int32 anymode = 0;
 
-float x_scale = 1, y_scale = 1;
 int32 x_offset = 0, y_offset = 0;
 int32 x_limit = 0, y_limit = 0;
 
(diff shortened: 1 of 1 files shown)


a740g (approved):

a740g:
> Because x_scale and y_scale are not used anywhere else I just removed them completely.

We can comment this out as well then?
Around 35227: `static float x_scale, y_scale;`

mkilgore:
> > Because x_scale and y_scale are not used anywhere else I just removed them completely.
> 
> We can comment this out as well then? Around 35227: `static float x_scale, y_scale;`

I did see those, I considered removing them as they're not used either. The catch is that because `x_scale` and `y_scale` are actually declared on that line those aren't the same `x_scale` and `y_scale` that I removed, they just have the same name, so I left them be,

I'm going to make a card around examining these variables and cleaning up some of the unused stuff and the resizing logic, there's more to do than just what I did here and @SteveMcNeill's been hitting similar issues today. Conceptually it doesn't seem like it should be super complicated, there's only so many things that can trigger a resize, so it seems like some decent progress could be made.

RhoSigma-QB64 (approved):

PR #164, opened 2022-09-05
Add MIDI support and `Compiler Settings` command line options

mkilgore:
This adds MIDI support to the language as a new unstable feature. There
are two new metacommands that come with this:

```vb
$Unstable: Midi
$MidiSoundFont: [Default|"filename"]
```

Once midi support is setup with those commands, you can `_SNDOPEN()` a MIDI
file and play it like any other sound file

The `$Unstable` command is required to be able to use any of the other
MIDI commands, and just signifies that this is not a full part of the
language yet and may change in breaking ways before the API is
finalized.

The `$MidiSoundFont` command enables MIDI support in the compiled program,
and also specifies what sound font to use to play MIDI files. `Default`
will make use of the soundfont placed at
'./internal/support/default_soundfont.sf2', and otherwise a filename can
be specified to use any soundfont wanted. In both cases, the selected sound
font is compiled into the executable and then loaded at runtime.

Note that MIDI is not supported when using the OpenAL backend, and we report an
error in that situation.

Command line parameters for setting things in the `Compiler Settings` dialog
were also added. This is actually a feature requested by a few different people
so it's nice that I finally got around to doing it, I needed it for some of the
automated testing for the MIDI stuff.

I would also like to callout some improvements I made to the testing system. I
added documentation in `docs/testing.md` regarding it, along with more complete
information on how to actually add new tests.


internal/c/parts/audio/miniaudio.h
```diff
@@ -61739,34 +61823,6 @@ MA_API ma_result ma_decoder_init_memory(const void* pData, size_t dataSize, cons
 #define MA_HAS_PATH_API
 #endif
 
-#if defined(MA_HAS_PATH_API)
```
a740g, line 61742: The rest of the changes looks good. But I am not sure if modifying miniaudio.h is a good idea. I have seen mackron pushing changes to the miniaudio dev branch. Some of these seem useful and important when they make it to release. If we want to use the newer versions when it is released, we would need to modify the header again to accommodate our changes. Which I think should be avoided. My recommendation would be to keep miniaudio.h intact and make all our changes outside of it. Looking at these changes, I am quite certain it is not too much we are changing... but still.

a740g (commented):
Everything looks good. However, we'll need to decide if we are ok making changes to external libraries directly as this will become an extra step when swapping out these libraries with newer versions in the future.

RhoSigma-QB64 (approved):

internal/c/parts/audio/miniaudio.h
mkilgore: Yeah I agree, I didn't really feel good about it at the time but we were using that internal stuff and I didn't want to completely rewrite it when it functions perfectly fine. I'll revert the changes to `miniaudio.h` and then just copy their code elsewhere (with new function names) for our use.

a740g (approved):

PR #165, opened 2022-09-05
Fix compiling source files with single quotes in name

mkilgore:
Using single quotes in a file name wasn't working because the Makefile
was using single quotes itself to quote the filename when providing it
to the compiler. The Makefile was switched to use double quotes instead,
which work just as well but are allowed to contain single quotes inside
of them.

Since double quotes are a legal character to have in a file name on
Linux and Mac OS, I added extra logic to start escaping double quote
characters.

I added a test to verify that we can compile source files with single
quotes in the name. I could not add a test for double quotes since such
files cannot exist on Windows and thus cannot easily be added to the
repository.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -386,11 +386,11 @@ clean: $(CLEAN_DEP_LIST)
 	$(RM) $(call FIXPATH,$(CLEAN_LIST))
 
 $(EXE): $(EXE_OBJS) $(EXE_LIBS)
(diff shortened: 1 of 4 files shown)


mkilgore:
I didn't catch that this needs some of the testing changes in #164, the build for this PR will be fixed after that PR gets merged.

RhoSigma-QB64 (approved):

a740g (approved):

Issue #170, opened 2022-09-08
Allow `SOUND` to play sounds without blocking the caller when `PLAY "MB` is used


a740g:
![image](https://user-images.githubusercontent.com/47574584/189196089-c15cfc3b-8137-4083-bed1-7dd11fc19aa3.png)

> Music (PLAY statement) and SOUND are set to run in the background. That is, each note or sound is placed in a buffer, allowing the BASIC program to continue executing while the note or sound plays in the background. The maximum number of notes that can be played in the background at one time is 32.

This is currently not implemented and can be easily implemented.

This will be completed after https://github.com/QB64-Phoenix-Edition/QB64pe/pull/164 is merged.

PR #171, opened 2022-09-09
Fix #169 & #170

### LINKED ISSUE #169: Change the miniaudio backend functions to return 0 as invalid handle instead of -1
See http://qb64phoenix.com/forum/showthread.php?tid=849&pid=6407#pid6407 for the complete background. This will also require the wiki to to updated so that all handle related content is consistent.

This will be completed after https://github.com/QB64-Phoenix-Edition/QB64pe/pull/164 is merged.

a740g:
This fixes #169 & #170.

Although the internal handle allocator still returns -1, this is caught and returned as 0 if allocation fails. All handle 0 operations were already being handled as a special case. Since handle 0 is internally allocated and used by `SOUND` & `PLAY`, this change does not break anything.

`PLAY` was already using a local variable called `mb` to keep track of `"MB"` & `"MF"`. We simply moved this variable to a global scope and allowed `SOUND` to use it as well. So, now when `PLAY "MB"` is used before any `SOUND` statements, it will allow sounds to play in the background. The user can set default behavior by simply specifying `PLAY "MF"`. This mirrors QuickBASIC 4.5 behavior.


mkilgore (approved):
It looks good to me, I think it would be worth adding a few tests though if you don't mind. Testing that `_SNDOPEN()` returns zero with an invalid filename should be pretty straight forward.

The `PLAY` and `SOUND` behavior are a little harder to test, but you should be able to do it by using `TIMER` to simply check how long a `SOUND` command takes after a `MB` and `MF`. You can just print whether the time was less than some threshold.

That said a test for the invalid handle seems much more important to me than the `SOUND` ones if you're not interested in doing those 😄

a740g:
> It looks good to me, I think it would be worth adding a few tests though if you don't mind. Testing that `_SNDOPEN()` returns zero with an invalid filename should be pretty straight forward.

I had actually planned for this but could not zero in on a place to put it under `tests`. 😁 Any suggestions?

a740g:
Merged without the new audio tests. These will be added later.

PR #180, opened 2022-09-17
Fix #175

### LINKED ISSUE #175: Allow loading 8bpp and lower color images without dithering the image and changing the palette
QB64-PE uses `stb_image` & `dr_pcx` libraries to load all image files when requests are made using `_LOADIMAGE`.

Both `stb_image` and `dr_pcx` automatically `de-palettizes` the image on load. So, there is no way to tell what kind of color information these images had after the image data is received from these libraries.

See https://github.com/nothings/stb/issues/510 and https://github.com/nothings/stb/issues/1124

So when the user loads a 24-bit image using `_LOADIMAGE("my_image.jpg", 256)`, then:

1. The image is converted to 32bpp by `stb_image` while loading
2. Once we have the 32bpp data, `_LOADIMAGE` converts the image to 8bpp and applies ordered dithering
3. An adaptive palette is created or the image is remapped to QB64-PE master palette (depending on the `_LOADIMAGE` `bpp` parameter)

This results in an acceptable 8-bit image quality.

However, when the user loads an 8-bit palettized image using `_LOADIMAGE("my_image.gif", 256)`, `stb_image` `de-palettizes` the image and the same steps are applied to the data (as noted above). Hence, there are two side-effects:

1. The color palette is not the same compared to the original image it was loaded from
2. Some artifacts may be seen in the final image due to the ordered dithering step

a740g:
@mkilgore @RhoSigma-QB64 

This fixes most of the issues mentioned in #175 thanks to the idea @RhoSigma-QB64 gave.

We now simply check the unique colors in the 32bpp image and extract the 8bpp image out of it if the number of unique colors are <= 256. This gives us a pixel perfect 256 color image without any kind of dithering. If the number of unique colors is > 256 then it simply does what it was before (which basically means that only true 32bpp images will get the dithering treatment).

One issue that remains and that cannot be resolved in any reasonable way (unless we use a different image library) is that the converted image's palette is rearranged and the indexes will be different compared to what is there in the loaded 8bpp image file. This can be an inconvenience for people who are expecting certain colors to be in certain indexes. This can however be solved on the BASIC side by remapping the image to a user palette. Which should be a 1:1 map.


internal/c/libqb/include/image.h
```diff
@@ -0,0 +1,58 @@
...
+//-----------------------------------------------------------------------------------------------------
+// HEADER FILES
+//-----------------------------------------------------------------------------------------------------
+#include <stdint.h>
+#include <stdio.h>
+//-----------------------------------------------------------------------------------------------------
+
+//-----------------------------------------------------------------------------------------------------
+// MACROS
+//-----------------------------------------------------------------------------------------------------
```
mkilgore, line 27: This is fine as-is, but I create #181 to look at improving our logging infrastructure and combining this with the similar audio logic you added.

internal/c/parts/video/image/image.cpp
```diff
@@ -0,0 +1,408 @@
...
+    }
+
+    int i, j;
+    auto uniqueColors = 0; // As long as this is <= 256 we will keep going until we are done
+    auto src32bpp = (uint32_t *)src;
+    for (i = 0; i < w * h; i++) {
+        auto srcColor = src32bpp[i];
+
+        // Check if the src color exists in our palette
+        for (j = 0; j < uniqueColors; j++) {
```
mkilgore, line 202: The logic goods good to me, I'm curious if you've seen how it performs with large images with lots of colors. I think it's fine even if it's a bit slower, if necessary I suspect we could optimize this a bit though.

mkilgore (approved):

internal/c/libqb/include/image.h
a740g: I am all in for this. We can make use of an existing library instead of writing our own. I'll research whenever I get some time.

internal/c/parts/video/image/image.cpp
a740g: Honestly, I did not think of this from a performance standpoint since all this happens only during image load. Also since `uniqueColors` is always `<= 256` I am not sure if it will have any significant slowdowns. Although, it could, if the images are unusually large. I'll run some tests and share here. We can review these numbers before merging the PR.

RhoSigma-QB64 (approved):

internal/c/parts/video/image/image.cpp
a740g: image
image

lol. We are taking a minute and 2 seconds to load 100 1080p images with all the conversion, compared to 5.8 seconds at 32bpp without any conversion. I am all ears. 😁

mkilgore: Ouch :(

I was thinking you could try using a hash table that maps the 32-bit colors to the 8-bit indexes picked for them. That would avoid the `O(n)` comparisons for each pixel by replacing them with a check of whether the 32-bit color is already in the hashmap or not.

a740g:
I am placing the QB64 test code (used above) here just in case we need to look at this stuff in the future.
```VB
DefLng A-Z
Option _Explicit
Option _ExplicitArray

Screen _NewImage(1280, 800, 32)
Dim handle As Long, fileName As String, t As Double, i As Long

fileName = "sample.gif"
t# = Timer

For i = 1 To 100
    handle = _LoadImage(fileName, 256)
    If handle > -2 Then
        Print "Failed to load "; fileName
        End
    End If
    _FreeImage handle
Next

Print Using "###.### seconds to load & 100 times."; Timer - t#, fileName

Sleep

handle = _LoadImage(fileName, 256)
_PutImage (0, 0), handle
_FreeImage handle

End
```

internal/c/parts/video/image/image.cpp
grymmjack: > Honestly, I did not think of this from a performance standpoint since all this happens only during image load. Also since `uniqueColors` is always `<= 256` I am not sure if it will have any significant slowdowns. Although, it could, if the images are unusually large. I'll run some tests and share here. We can review these numbers before merging the PR.

I would only be using smaller modest 256 color images myself. 320x200 - 1024x768 tops

mkilgore: > I would only be using smaller modest 256 color images myself. 320x200 - 1024x768 tops

The good news is that since this PR moved the code into a separate `cpp` file we realized we're able to always build this portion of the code with `-O2` even though we haven't enabled it for the rest of the code. I forget the exact numbers @a740g gave me but I believe after the addition of the hashmap and using `-O2` it only takes 8 seconds to load those images, so significantly better :)

grymmjack: That's great. Thanks @mkilgore and @a740g

PR #183, opened 2022-09-20
Update setup_lnx.sh

AtomicSlaughter:
Changed setup script to remove make and g++ to avoid compile errors, with make.


mkilgore:
@AtomicSlaughter Hey :) This change looks good, but can you remove the commits modifying `ci.yml`? You shouldn't need to touch it, the builds aren't running because a maintainer needs to approve them to run. I'm working at the moment though so I won't be able to give this a closer look till later.

Issue #195, opened 2022-09-26
_BLUE does not work in CONST


mkilgore:
`_RED` and `_GREEN` work fine, but trying to use `_BLUE` produces an error.

I believe the issue is that in `EvaluateNumber$` there's special handling for functions with more than one parameter (to allow commas), and `_BLUE` is missing from that `CASE` entry. Adding it lets `_BLUE` work correctly.

Issue #194, opened 2022-09-26
_RGBA does not work in CONST


mkilgore:
It's intended to work but doesn't, it just produces an error when you attempt to use it such as `CONST val = _RGBA(1, 2, 3, 4, 2)`

It appears to be an ordering issue, in the list of operators `_RGB` is listed before `_RGBA`, and since the names match at the beginning `Evaluate_Expression$` thinks that we're calling `_RGB` with an unnecessary `A` after it (which triggers an error).

Issue #177, opened 2022-09-13
Undeclared consts not caught in Option Explicit


SpriggsySpriggs:
```
Option _Explicit

Const TFL_WALK = &H00000002
Const TFL_CROUCH = &H00000004
Const TFL_BARRIERJUMP = &H00000008
Const TFL_JUMP = &H00000010
Const TFL_LADDER = &H00000020
Const TFL_WALKOFFLEDGE = &H00000080
Const TFL_SWIM = &H00000100
Const TFL_WATERJUMP = &H00000200
Const TFL_TELEPORT = &H00000400
Const TFL_ELEVATOR = &H00000800
Const TFL_JUMPPAD = &H00040000
Const TFL_AIR = &H00080000
Const TFL_WATER = &H00100000
Const TFL_FUNCBOB = &H01000000

Const TFL_DEFAULT = TFL_WALK Or TFL_CROUCH Or TFL_BARRIERJUMP Or TFL_JUMP Or TFL_LADDER Or TFL_WALKOFFLEDGE Or TFL_SWIM Or TFL_WATERJUMP Or TFL_TELEPORT Or TFL_ELEVATOR Or TFL_AIR Or TFL_WATER Or TFL_JUMPPAD Or TFL_FUNCBOB

```
If you erase part of the letters from the constants that are in the TFL_DEFAULT declaration (not in their individual declarations), you can see that it still gives an "OK" in the IDE message line. I tried it with several different constants in that expression and it still gave an "OK". With Option _Explicit enabled, it would be expected that the constant evaluator would catch constants that are not yet declared. It looks like constant declarations with fewer constants are caught but the message line says it is a compiler error rather than an undeclared variable/const.
![image](https://user-images.githubusercontent.com/10662502/190025412-ead65665-9c38-488e-84b6-ca1bbfc276bc.png)

mkilgore:
Messing around with it, it's pretty clearly more broken than just an issue with `Option _Explicit`, you get nonsense values when using non-existent `CONST` variables.

From what I've worked out, variable names which aren't resolved to existing `CONST` variables don't get filtered out by `Evaluate_Expression$()`. The bad names then get passed to `evaluateconst$()`, and depending on how exactly you write your expression it will incorrectly pass the name of the variable you wrote to `_CV(_INTEGER64, constvar$)`. The resulting value gets used in place of the variable name in your expression, completely screwing up the actual value.

[...]

Issue #196, opened 2022-09-26
CONST applies negation higher than exponentiation


mkilgore:
CONST evaluates `-2 ^ 2` as `(-2) ^ 2` and thus gives 4. However, if you actually evaluate this expression in Ex. a `PRINT` statement you'll find it evaluates to `-4` because QB64's order of operations treat that as `-(2 ^ 2)`, exponentiation is higher then negation.

I'm considering this a bug since `CONST` should have the same operator order as regular QB64 (IE. Any expression evaluated as `CONST` should give exactly the same value as evaluating it as a non-`CONST`), but it's notable that exponents are not allowed to be in `CONST` expressions in QB45 for some reason, so this is not a compatibility issue with QB45.

PR #197, opened 2022-09-26
Fix various CONST issues and add CONST testsuite

mkilgore:
While working on fixing #177 that @SpriggsySpriggs ran into I created a testsuite for CONST. Doing this exposed a variety of issues, a few of them (#194, #195) are fixed in this PR, and others (#192, #193, #196) are not. I suspect we may want to rewrite some of the underlying expression parsing logic, the play between `evaluateconst$` and `Evaluate_Expression$` is quite odd, they basically repeat some of the same logic and all the better error messages from `Evaluate_Expression$` get thrown out. Additionally I've written expression parsers before and there are some fairly nice algorithms for it that make it much easier and would likely fix all of the issues I found. That said, these changes at least address some of the issues and the testsuite should serve as a way of verifying the behavior of any `CONST` changes we make in the future :)

Brief descriptions of the issued I fixed:

1. #177 - `evaluateconst$` wasn't checking that the arguments to a boolean operator had proper types, so undefined variable names just get treated as integers (meaning, the bytes making up their name gets treated as an integer). We already have the type information so I just added two checks that error if either parameter doesn't have a type.
  a. Note that `Option _Explicit` doesn't actually matter here, even QB45 errors if you try to use undefined variable names in `CONST` expressions.
3. #194 - `_RGBA` didn't work because the parser thought it was `_RGB` by mistake. Changing the order the functions are listed in fixed this, because now `_RGBA` gets matched first.
4. #195 - `_BLUE` didn't work because it wasn't included in a list of functions that accept more than one parameter. Adding it to the list fixes it.

Fixes: #177
Fixes: #194 
Fixes: #195


RhoSigma-QB64 (approved):

mkilgore:
For reference, I resolved the test issues by allowing the floating point values to be within an expected range instead of an exact  value. There might be an argument to say that the calculated floating point constants should always be identical regardless of platform, but we can consider and look into that at a later time :shrug:

PR #201, opened 2022-09-30
Fix compiling bas files relative to the CWD

mkilgore:
Fix was pulled from QB64Official/qb64#17 by @DualBrain, I just added tests around it.

Most (all?) compilers allow you to run the compiler from a separate directory than the compiler itself is located in and compile source files relative to that directory. QB64-PE however does not allow that, for a variety of reasons it always search for the provided source file relative to the location of the QB64-PE compiler rather than the CWD it was run from. This is pretty unexpected behavior in a lot of cases, and also doesn't give very helpful error messages either.

This change has us check if the source file exists at the given CWD location, and if it does we will prepend the CWD to produce the correct path to the file.

To test that this behavior works as expected I modified `compile_test.sh` to compile from within the test directories using a relative path directly to the test file, this fails with current QB64-PE versions because it can't find the source file.

Additionally, I was unsure of whether this would impact the behavior of `'$include`, so I added some tests around include that uses various combinations of paths relative to QB64-PE and relative to the source file being compiled, and they all find the files as expected so I think it's fine.

### DIFF
--- a/tests/compile_tests.sh
+++ b/tests/compile_tests.sh
@@ -60,7 +60,7 @@ do
     cd "./tests/compile_tests/$category"
 
     # -m and -q make sure that we get predictable results
-    "../../../$QB64" $compilerFlags -m -q -x "$testName.bas" -o "$EXE" 1>"../../../$compileResultOutput"
+    "../../../$QB64" $compilerFlags -q -m -x "$testName.bas" -o "../../../$EXE" 1>"../../../$compileResultOutput"
     ERR=$?
 
     popd >/dev/null
--- a/tests/compile_tests.sh
+++ b/tests/compile_tests.sh
@@ -56,9 +56,14 @@ do
         compilerFlags=$(cat "./tests/compile_tests/$category/$testName.flags")
     fi
 
+    pushd . >/dev/null
+    cd "./tests/compile_tests/$category"
+
     # -m and -q make sure that we get predictable results
-    "$QB64" $compilerFlags -m -q -x "./tests/compile_tests/$category/$testName.bas" -o "$EXE" 1>"$compileResultOutput"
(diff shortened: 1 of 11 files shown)


RhoSigma-QB64 (approved):

tests/compile_tests.sh
```diff
@@ -56,9 +56,14 @@ do
         compilerFlags=$(cat "./tests/compile_tests/$category/$testName.flags")
     fi
 
+    pushd . >/dev/null
+    cd "./tests/compile_tests/$category"
+
     # -m and -q make sure that we get predictable results
-    "$QB64" $compilerFlags -m -q -x "./tests/compile_tests/$category/$testName.bas" -o "$EXE" 1>"$compileResultOutput"
+    "../../../$QB64" $compilerFlags -m -q -x "$testName.bas" -o "$EXE" 1>"../../../$compileResultOutput"
```
mkilgore, line 63: Taking a closer look the `-o` flag might be a problem, I didn't notice but it's still relative to the location of qb64pe when using it. Unfortunately unlike with selecting the source file I'm not sure there's a simple non-breaking solution for this one, I'll have to think about it. We could add a new flag for it but that's a bit ugly...

mkilgore: After some thinking I we can fix it in a non-breaking way, we can just make it so that the `-o` location is relative to the same location the source file was found relative too. I'll get that done in this PR and we can review it again.

My logic is that it's not a breaking change because currently we only allowed source files relative to `qb64pe`, and in that situation we'll treat the `-o` location as relative to `qb64pe`, the same as it currently works. In the new situation of a source file that's relative to the cwd/`_STARTDIR$` that just gave an error before so it has no defined behavior, making `-o` relative to the source location thus doesn't break any previously working usages. Additionally I think it's by far the expected behavior, so it makes sense to make that change.

RhoSigma-QB64 (approved):

Issue #185, opened 2022-09-21
Improve `_SNDPLAYCOPY`


a740g:
`_SNDPLAYCOPY handle&[, volume!]` in it's current state is useful but limiting. This can be used to quickly play a copy of a loaded sound and playing the sound at a certain volume.

However, if one wants to apply stereo panning or a 3D position, then it is impossible to do so. To achieve this one must use `_SNDCOPY` followed by `_SNDVOL` and / or `_SNDBAL`.

To improve the usefulness of `_SNDPLAYCOPY`, it can be extended as follows:
`_SNDPLAYCOPY handle&[, volume!][, x!][, y!][, z!]`

This will allow users to apply volume, panning and position while playing sound copies without breaking existing functionality.

PR #208, opened 2022-10-04
Implement `_SNDPLAYCOPY` enhancement #185

a740g:
This implements enhancements to `_SNDPLAYCOPY` as described in #185.

The extra `[, x!][, y!][, z!]` parameters are optional and hence this does not have any breaking changes.
The extra parameters can also be used with the old OpenAL audio backend. However, it does not change the properties of the sound.

We could have backported this feature to the OpenAL backend. However, considering that it is on life-support and may be removed it the future, I do not see any incentive.

This also corrects a minor typo in the help text for `_SNDBAL` that I discovered while implementing the changes in `subs_functions.bas`.

### DIFF
--- a/internal/c/parts/audio/audio.cpp
+++ b/internal/c/parts/audio/audio.cpp
@@ -1523,8 +1523,11 @@ void sub__sndplay(int32_t handle) {
 /// </summary>
 /// <param name="handle">A sound handle to copy</param>
 /// <param name="volume">The volume at which the sound should be played (0.0 - 1.0)</param>
+/// <param name="x">x distance values go from left (negative) to right (positive)</param>
+/// <param name="y">y distance values go from below (negative) to above (positive).</param>
+/// <param name="z">z distance values go from behind (negative) to in front (positive).</param>
 /// <param name="passed">How many parameters were passed?</param>
-void sub__sndplaycopy(int32_t src_handle, double volume, int32_t passed) {
+void sub__sndplaycopy(int32_t src_handle, double volume, double x, double y, double z, int32_t passed) {
     // We are simply going to use sndcopy, then setup some stuff like volume and autokill and then use sndplay
     // We are not checking if the audio engine was initialized because if not we'll get an invalid handle anyway
     int32_t dst_handle = func__sndcopy(src_handle);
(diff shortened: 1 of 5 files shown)


internal/c/qbx.cpp
```diff
@@ -216,11 +217,6 @@ extern void sub__maptriangle(int32 cull_options, float sx1, float sy1,
                              int32 smooth_options, int32 passed);
 extern void sub__depthbuffer(int32 options, int32 dst, int32 passed);
 extern int32 func_play(int32 ignore);
-extern int32 func__sndrate();
```
mkilgore, line 219: This is some good clean-up, thanks for doing that :)

internal/c/parts/audio/out/src.c
```diff
@@ -1149,15 +1149,15 @@ void sub__sndclose(int32 handle) {
 
 //"macros"
 
-void sub__sndplaycopy(int32 handle, double volume, int32 passed) {
+void sub__sndplaycopy(int32 handle, double volume, double x, double y, double z, int32 passed) {
```
mkilgore, line 1152: Yeah I agree with not bothering with adding support in the OpenAL backend. We should create a plan for getting rid of it, I'll look into how we can organize that.

internal/c/parts/audio/audio.cpp
```diff
@@ -1523,18 +1523,39 @@ void sub__sndplay(int32_t handle) {
...
     int32_t dst_handle = func__sndcopy(src_handle);
 
     // Check if we succeeded and then proceed
     if (dst_handle > 0) {
         // Set the volume if requested
-        if (passed)
+        if (passed & 1)
             ma_sound_set_volume(&audioEngine.soundHandles[dst_handle]->maSound, volume);
 
+        if (passed & 4 || passed & 8) {                                                                   // If y or z or both are passed
```
mkilgore, line 1541: Is it an option to just call `sub__sndbal()` here? Since we're basically just duplicating the logic.

RhoSigma-QB64 (approved):

mkilgore (approved):

internal/c/parts/audio/audio.cpp
a740g: I was actually tempted to just call `sub__sndbal` and be done with it. However, I realized that we were going to set balance just once for the sound copy and the sound would eventually get destroyed. So, we do not need to care about what the old `x, y, z` was like `sub__sndbal` does and these are just set to zero if they are not passed. Also, `x` is completely optional unlike `sub__sndbal`.

Issue #189, opened 2022-09-22
Platform agnostic `common dialogs` support


a740g:
We should provide a way to support `common dialogs` (like MessageBox, InputBox, PasswordBox, ColorPicker, OpenFileDialog, SaveFileDialog, SelectFolderDialog, Tray-popup, Notification Beep) in a platform agnostic manner.

`libqb.cpp` already has rudimentary support for `MessageBox` & `Alert`. However, these can be significantly improved and extended.

From my observations most folks resort to calling native OS libraries for these functionalities which ties the code to a particular platform. Also, supporting common dialogs for multiple platforms on the BASIC side is rather complicated and complex.

My recommendation is to use a tried, tested and compact library like [tiny file dialogs](https://sourceforge.net/projects/tinyfiledialogs/).
Using this we can then have VB-style dialog functions like `_GetOpenFilename`, `_MsgBox`, `_InputBox`, etc...

SpriggsySpriggs:
I like this idea. I don't typically do cross-platform code so when I do dialogs they always are platform specific. However, I could cut down on BAS code by having an internal dialog function like this.

PR #216, opened 2022-10-23
This implements cross-platform common dialogs for QB64-PE (issue #189)

a740g:
The following subroutines and functions are implemented. These are modeled to closely resemble [tiny file dialogs](https://sourceforge.net/projects/tinyfiledialogs/) and the common dialog features in VBDOS 1.0.

```VB
SUB _NOTIFYPOPUP [title$][, message$][, iconType$]
SUB _MESSAGEBOX [title$][, message$][, iconType$]
FUNCTION _SELECTFOLDERDIALOG$(title$[, defaultPath$])
FUNCTION _COLORCHOOSERDIALOG&(title$[, defaultRGB&])
FUNCTION _MESSAGEBOX&(title$, message$, dialogType$, iconType$[, defaultButton&])
FUNCTION _INPUTBOX$(title$, message$[, defaultInput$])
FUNCTION _OPENFILEDIALOG$(title$, defaultPathAndFile$, filterPatterns$, singleFilterDescription$[, allowMultipleSelects&])
FUNCTION _SAVEFILEDIALOG$(title$, defaultPathAndFile$, filterPatterns$, singleFilterDescription$)
```

This PR also cleans up `alert()`, `MessageBox()` & `MessageBox2()` implementations in `libqb` and replaces those with `tiny file dialogs`. Thus, `tiny file dialogs` becomes a required dependency.

This also includes minor changes to consolidate things in the right place. For example, `qbs_new_txt_len()` is now declared in `libqb.h`. Some commonly used macros like the `IMAGE_...` stuff has now been moved to appropriate header files in `libqb/include`.


SpriggsySpriggs:
Wow! ❤️

RhoSigma-QB64:
@a740g Don't forget the Wiki pages for those new SUBs and FUNCTIONs.

internal/c/parts/video/image/image.cpp
```diff
@@ -29,7 +29,7 @@
 // We'll likely keep the 'include' this way because I do not want to duplicate stuff and cause issues
 // Matt is already doing work to separate and modularize libqb
 // So, this will be replaced with relevant stuff once that work is done
-#include "../../libqb.h"
+#include "../../../libqb.h"
```
mkilgore, line 32: Huh, it seems like this should have failed to compile if the path was wrong 😮

source/subs_functions/subs_functions.bas
```diff
@@ -1970,7 +1970,7 @@ id.args = 2
 id.arg = MKL$(STRINGTYPE - ISPOINTER) + MKL$(STRINGTYPE - ISPOINTER)
 id.specialformat = "?[,?]"
 id.ret = ULONGTYPE - ISPOINTER
-id.hr_syntax = "_SNDOPEN(fileName$)"
+id.hr_syntax = "_SNDOPEN(fileName$[, capabilities$])"
```
mkilgore, line 1973: Sneaking this one in I see 😆

licenses/license_tinyfiledialogs.txt
```diff
@@ -0,0 +1,47 @@
+--------------------------------------------------------------------------------
```
mkilgore, line 1: Would you mind adding this to the [licenses/README.md](https://github.com/QB64-Phoenix-Edition/QB64pe/blob/main/licenses/README.md) in the `QB64 Phoenix Edition Runtime` table?

internal/c/parts/gui/tinyfiledialogs.h
```diff
@@ -0,0 +1,303 @@
...
+- On unix: it tries command line calls, so no such need (NO LINKING).
+- On unix you need one of the following:
+  applescript, kdialog, zenity, matedialog, shellementary, qarma, yad,
+  python (2 or 3)/tkinter/python-dbus (optional), Xdialog
+  or curses dialogs (opens terminal if running without console).
+- One of those is already included on most (if not all) desktops.
+- In the absence of those it will use gdialog, gxmessage or whiptail
+  with a textinputbox. If nothing is found, it switches to basic console input,
+  it opens a console if needed (requires xterm + bash).
+- for curses dialogs you must set tinyfd_allowCursesDialogs=1
```
mkilgore, line 287: I mentioned it in the discord but I think we should really consider enabling this for `$CONSOLE` programs. The comments in the code suggest that we can either enable it as a fallback in the event the graphics-based dialogs don't work, or always force the console-based dialog, and also I think we can more or less change it at any time.

I'm not entirely sure yet, but perhaps we should determine whether the console-based dialog is allowed/forced based on whether `_CONSOLE ON` is currently enabled. That's always on for `$CONSOLE:ONLY`, and would optionally be on for programs like QB64-PE which only use console output in some circumstances. I think that gives the most natural result, as it means using something like `qb64pe` gives regular graphical dialogs like expected, but using `qb64pe -x` will always give console-based dialogs if they come up for some reason (Ex. error).

Regardless of the exact semantics, allowing console dialogs would be nice because it lets the user see error pop-up's or these dialogs even if their program is console-based and there is no graphics server running (which is common on Linux).

internal/c/parts/gui/gui.cpp
```diff
@@ -0,0 +1,365 @@
...
+    qbs_set(aTitle, qbs_add(qbsTitle, qbs_new_txt_len("\0", 1)));
+
+    if (passed)
+        qbs_set(aDefaultPath, qbs_add(qbsDefaultPath, qbs_new_txt_len("\0", 1)));
+    else
+        qbs_set(aDefaultPath, qbs_new_txt_len("\0", 1));
+
+    auto sFolder = tinyfd_selectFolderDialog((const char *)aTitle->chr, (const char *)aDefaultPath->chr);
+
+    return qbs_new_txt(sFolder);
```
mkilgore, line 249: This seems like it could be a potential problem. I assume `sFolder` is backed by a static buffer in `tinyfd`. We probably want to duplicate it here so that the returned `qbs` doesn't randomly change if the user calls `_SelectFolderDialog` a second time. This goes for some of the other functions as well.

Admittedly I don't understand `qbs` well enough to know exactly how to do this. It looks like other functions that return strings use `qbs_new()` to create a new string and then copy the contents in, presumably that would work here too 🤷

internal/c/parts/video/image/image.cpp
a740g: Honestly, I don't know how it was compiling before. XD

Makefile
```diff
@@ -215,6 +216,15 @@ else
 	QBLIB_NAME := $(addsuffix 0,$(QBLIB_NAME))
 endif
 
+ifneq ($(filter y,$(DEP_COMMON_DIALOGS)),)
+	CXXFLAGS += -DDEPENDENCY_COMMON_DIALOGS
```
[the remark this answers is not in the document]
a740g: Ok. That makes sense. I'll remove this.

internal/c/libqb.cpp
```diff
@@ -501,6 +501,21 @@ int32 dont_call_sub_gl = 0;
 
 void GLUT_DISPLAY_REQUEST();
 
+// a740g: These are driven via tiny file dialogs
```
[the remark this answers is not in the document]
a740g: Ok. That would simplify a lot of things then and we also would not need the additional makefile logic. Makes sense to me.

internal/c/parts/gui/tinyfiledialogs.h
a740g: Ok. We can set `tinyfd_allowCursesDialogs=1` when `$CONSOLE:ONLY` is specified on Linux. Or should we do it for all platforms? There is a `tinyfd_forceConsole` also. I'll also try a couple of things and share my findings.

internal/c/parts/gui/gui.cpp
a740g: That is a good catch. I was assuming that `qbs_new_txt` and `qbs_new_txt_len` makes copies of the strings. I looked at those closely and it seems those aren't. They simply do a `newstr->chr = (uint8 *)txt`. `func_string` does what you are saying.  I'll rewrite this and the others.

internal/c/parts/gui/tinyfiledialogs.h
mkilgore: I think it depends how well the console dialogs work, I suspect the Linux ones are better than the Windows ones. Console-only Windows (no graphics at all) is also significantly less common than for Linux (in my experience anyway), so it's probably less of a concern for Windows.

We'll probably want to do a bit with separate `cpp`s to initialize that setting without using `#ifdef`s in `libqb.cpp`, after we determine how we want it to function I could help with that. Also we could potentially leave this for another card if you want.

internal/c/parts/gui/gui.cpp
mkilgore: Yeah I noticed the behavior of `qbs_new_txt` a little while back, it seems designed for creating `qbs`s out of static strings, which I suppose makes sense. As part of the `qbs` refactor ideally we'll document some of these functions 😆...

Makefile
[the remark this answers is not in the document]
a740g: All DEP_COMMON_DIALOGS and DEPENDENCY_COMMON_DIALOGS removed from makefiles and sources.

internal/c/libqb.cpp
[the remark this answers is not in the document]
a740g: Moved these to gui.cpp

licenses/license_tinyfiledialogs.txt
a740g: Done. Please check the lines added though.

internal/c/parts/gui/tinyfiledialogs.h
a740g: I am going to leave this for you. 😁 I cannot call myself any level of expert on Linux. I think we should check the behavior first though. I have no (or don't know a) way of testing this currently. I want to know what tfd does if we run `$CONSOLE:ONLY` stuff on Linux without X.

internal/c/parts/gui/gui.cpp
a740g: This is done. I modified the functions to return a `qbs` like `func_string` does.

Makefile
```diff
@@ -359,7 +360,7 @@ ifeq ($(OS),win)
 
 		LICENSE_IN_USE := $(filter-out freeglut,$(LICENSE_IN_USE))
 	else
-		CXXLIBS += -mwindows -lopengl32 -lglu32 -lwinmm
+		CXXLIBS += -mwindows -lopengl32 -lglu32 -lwinmm -lcomdlg32 -lole32
```
mkilgore, line 363: I think these guys are unnecessary :)

mkilgore (approved):

Makefile
a740g: Yeah. I noticed that too late. I already made the changes but was waiting for the checks to finish.

internal/c/parts/gui/tinyfiledialogs.h
mkilgore: Done, I made #219 for it :)

PR #236, opened 2022-11-05
Fix quoting of arguments to dialog functions on Linux

mkilgore:
tinyfiledialogs had lots of issue surrounding the quoting of the arguments to the dialog functions. The arguments are effectively placed verbatim into shell commands, and without any proper treatment they would get accidentally evaluated by the shell, which is a big problem. Additionally, any arguments containing quotes would just not work since they would screw up the underlying shell command.

This change fixes all that for Linux. Every argument to the dialog functions is now surrounded in single quotes, which prevents any evaluation of the string by the shell. Additionally, because this is standard the only characters that need special treatment are the single quote characters, and proper escaping for them has been added. The result is that the dialog functions can now accept arguments that have single and double quotes in them (on Linux).

We sadly don't have any automated testing for this functionality yet, but I manually tested all the dialog options I could (Xdialog, kdialog, yad, zenity, dialog, console, xmessage, etc...). Note that as part of this change I tossed the Python and perl based dialogs. I'm not familiar enough with either of those languages to know what kind of shenanigans we would have to do to make the string insertions into that code safe. For both of them I think it would be fairly rare that the required libraries and such would be on a given system so IMO it's unlikely they would get used anyway.

The `FIXME` entries that I left in the code are notes for Mac OS. I didn't touch that code because I couldn't yet test it, but I suspect we just need to escape the double quotes and we're good (which is easy enough). We might need to escape them twice, we'll see when we get there.

### DIFF
--- a/internal/c/parts/gui/tinyfiledialogs.c
+++ b/internal/c/parts/gui/tinyfiledialogs.c
@@ -170,7 +170,6 @@ char tinyfd_needs[] = "\
 \ntiny file dialogs on UNIX needs:\
 \n   applescript or kdialog or yad or Xdialog\
 \nor zenity (or matedialog or shellementary or qarma)\
-\nor python (2 or 3) + tkinter + python-dbus (optional)\
 \nor dialog (opens console if needed) ** Disabled by default **/\
 \nor xterm + bash (opens console for basic input)\
 \nor existing console for basic input";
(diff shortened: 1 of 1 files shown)


GeorgeMcGinn:
I can test this on a MAC if needed.

mkilgore:
> I can test this on a MAC if needed.

I'm pretty sure none of the changes I made here impact Mac OS since it should always use the applescript option (which I haven't fixed yet). I did remove the check for quotes though, so you could attempt to use those in arguments to the dialogs but I would expect them to just be broken for the moment.

That said I may take you up on that when I do attempt the applescript fixes soon :smile:

a740g (approved):

RhoSigma-QB64 (approved):

PR #237, opened 2022-11-06
Swap NotifyPopup to use Shell_NotifyIcon

mkilgore:
Fairly straight forward, the Powershell code was fundimentally just calling Shell_NotifyIcon under the covers which we can easily do ourself. The notification has to be tied to a window, so I simply create a hidden one to use.

A nice improvement here is that instead of having to use the Powershell icon we can make use of the EXE icon if one was provided. If one is not provided then we use the default 'application' icon. Additionally quotes can be in the message and title now with no issues.

Ex:
image

### DIFF
--- a/internal/c/parts/gui/tinyfiledialogs.c
+++ b/internal/c/parts/gui/tinyfiledialogs.c
@@ -1175,6 +1175,120 @@ int tinyfd_messageBoxW(
         }
 }
 
+static LRESULT CALLBACK hiddenNotifyWindowProc(HWND hwnd, UINT message, WPARAM wParam, LPARAM lParam)
+{
+    switch (message)
+    {
+    case WM_CREATE:
+        break;
+    case WM_DESTROY:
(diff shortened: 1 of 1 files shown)


a740g (approved):

RhoSigma-QB64 (approved):

SpriggsySpriggs:
Sweeeeet

mkilgore:
> Sweeeeet

Thanks for the tip about `Shell_NotifyIcon` BTW :) Initially I thought the notifications were different, but eventually I realized the Powershell script was just calling `Shell_NotifyIcon` under the covers 🤦

a740g:
> Thanks for the tip about `Shell_NotifyIcon` BTW :) Initially I thought the notifications were different, but eventually I realized the Powershell script was just calling `Shell_NotifyIcon` under the covers 🤦

Well merge it already. This is good stuff! 😁

PR #238, opened 2022-11-07
Fix _InputBox$ on Windows to use DialogBoxIndirect, and a few other changes

mkilgore:
This removes the VBS based InputBox for Windows and replaces it with a
version that uses DialogBoxIndirect to create the dialog. While it is a
bit more complicated in some respects, it removes any concerns about the
contents of the strings as they're no longer being inserted into the
generated script. It also has the advantage that it doesn't spawn
another process (which then shows up in the task bar in some situations).

With this changes quote characters are allowed in all of the parameters.
Additionally this is the last set of changes for Windows, Mac OS is the only
platform left.

This additionally contains the change caught by @a740g to mark the `lBuff`
array `static` in `tinyfd_openFileDialog()`. I also went ahead and removed
the last of the quote checks (which weren't doing anything on Windows
beyond keeping parity with the functionality for other platforms).

### DIFF
--- a/internal/c/parts/gui/tinyfiledialogs.c
+++ b/internal/c/parts/gui/tinyfiledialogs.c
@@ -1100,9 +1100,6 @@ int tinyfd_messageBoxW(
 
         if (aTitle&&!wcscmp(aTitle, L"tinyfd_query")){ strcpy(tinyfd_response, "windows_wchar"); return 1; }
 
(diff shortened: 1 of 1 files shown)


a740g (approved):

mkilgore:
I'm going to fix it to trigger a hidden/password input when the default text is blank and then I'll merge this.

PR #241, opened 2022-11-08
Fix PAINT when border color is not supplied (QB64Official/qb64#38)

mkilgore:
The border color parameter to PAINT is optional, but sub_paint was not
handling that case. What it should do in that situation is keep painting
until it finds pixels that are not the same color as the original
starting pixel was. Instead it would simply assume border color was
valid and paint until it finds color zero (the default parameter value
when it's not provided).

This was originally reported in QB64Official/qb64#2, and Walt
(@TheJoyfulProgrammer) fixed it in QB64Official/qb64#38. Functionally
this is the same change, however I'm checking `passed & 4` to see
whether a border color was provided rather than `bordercol == NULL`.
The later has problems if the provided bordercolor is zero, which is
allowed since zero is a valid color. The `passed` argument indicates
which of the function arguments were actually provided in the QB64
source that called PAINT.

Additionally, along with the `while (true)` loop that Walt changed I
went ahead and removed the duplication of the sections for each
direction. We now just use a couple of arrays to determine which
direction we're checking and loop over all 4.

This change introduces new image-based tests, where the expected resulting
image is checked into this repo and the test runs a variety of commands
to manipulate a new image in memory and then compare that against the
expected result in the repo. The created image is saved in the test results
so it can be examined in the event of a failure.

For reference, the `tile_border_white_bg` test is the one that fails with Walt's original change.


a740g (approved):

SteveMcNeill (approved):

TheJoyfulProgrammer:
> The border color parameter to PAINT is optional, but sub_paint was not handling that case. What it should do in that situation is keep painting until it finds pixels that are not the same color as the original starting pixel was. Instead it would simply assume border color was valid and paint until it finds color zero (the default parameter value when it's not provided).
> 
> This was originally reported in [QB64Official/qb64#2](https://github.com/QB64Official/qb64/issues/2), and Walt (@TheJoyfulProgrammer) fixed it in [QB64Official/qb64#38](https://github.com/QB64Official/qb64/pull/38). Functionally this is the same change, however I'm checking `passed & 4` to see whether a border color was provided rather than `bordercol == NULL`. The later has problems if the provided bordercolor is zero, which is allowed since zero is a valid color. The `passed` argument indicates which of the function arguments were actually provided in the QB64 source that called PAINT.
> 
> Additionally, along with the `while (true)` loop that Walt changed I went ahead and removed the duplication of the sections for each direction. We now just use a couple of arrays to determine which direction we're checking and loop over all 4.
> 
> This change introduces new image-based tests, where the expected resulting image is checked into this repo and the test runs a variety of commands to manipulate a new image in memory and then compare that against the expected result in the repo. The created image is saved in the test results so it can be examined in the event of a failure.
> 
> For reference, the `tile_border_white_bg` test is the one that fails with Walt's original change.

Please provide the source code to the "tile_border_white_bg" so I can verify it not working with the code I modified.

In C++ **NULL != 0**.

I did various tests with the border color equal to 0 with my modified code, and nothing failed. I need to see the code that failed to investigate further.

Thank you.

TheJoyfulProgrammer:
Using the TILEDEMO.BAS that DualBrain provided which showed the pattern-fill initially not working, here's a modified version with a white background with black line art and no fill for demonstration purposes.

Screenshot_20221110_025405

Here's the source code to the above screenshot:

Screenshot_20221110_030319

When I removed the remark ticks to the PAINT statements, which has no border color defined, here's what I get:

Screenshot_20221110_025556

Here's the source code:

Screenshot_20221110_025813

When I added a border color of '0', the pattern fills worked the same. Here's the source code to it:

Screenshot_20221110_025837

This is why I need the source code to the one you claimed to fail with my code.

mkilgore:
> Please provide the source code to the "tile_border_white_bg" so I can verify it not working with the code I modified.
> 
> In C++ **NULL != 0**.
> 
> I did various tests with the border color equal to 0 with my modified code, and nothing failed. I need to see the code that failed to investigate further.
> 
> Thank you.

`bordercol` is just an `int32`, it's not a pointer type. When you compare `NULL` against an integer it does just get treated as zero since that's the null-pointer-constant in C. Additionally because C++ doesn't allow implicit casts of `void *` it has to be defined as `#define NULL 0` and not `#define NULL (void *)0`.

You may be thinking of `nullptr` since that fixes this kind of issue and will not compare successfully against an integer (but is still allowed to be assigned to any pointer type without a cast). A quick C++ example you could try:

```cpp
printf("NULL equals zero: %d\n", 0 == NULL); // Prints 1
printf("nullptr equals zero: %d\n", 0 == nullptr); // Does not compile
```

For the test case I should have clarified, the test is in this PR's changes as it runs as part of CI. Here is a slightly modified version of the test that you can run directly (it omits the image comparison logic that CI does):

```vb
TILE$ = MKL$(&HAAFFBB55)

' Paint with tiling, black border color.
' background is colored white, circle is black.
'
' Note the starting location for Paint is neither
' white or black, the starting color should not
' matter when using a border color.
'
' Result should fill the entire circle with pattern
Screen 9

' Make the entire image white
Line (0, 0)-(127, 49), 7, BF

Circle (64, 25), 25, 0
PSet (64, 25), 15 ' Set the starting location to a third color

Paint (64, 25), TILE$, 0
```

The expected result is this:

image

PR #242, opened 2022-11-11
Swap to New Load Dialog

SteveMcNeill:
Without IDE toggle yet between old and new dialogs, for testing purposes and comments.


a740g:
`source/qb64pe.bas: 13,839 additions, 13,837 deletions not shown because the diff is too large. Please use a local Git client to view these changes.`
@SteveMcNeill I think you turned off the `Keywords in CAPTITALS` option unchecked by accident. :)

SteveMcNeill:
> 

Aye.  QB64 doesn't come with the "Keywords in CAPITALS" turned on, so I just didn't think about clicking that option. Maybe we should start shipping with that option clicked on so I won't do this with every Xth commit.  ;)

a740g:
> Maybe we should start shipping with that option clicked on so I won't do this with every Xth commit. ;)

I agree with you on this one. Eventhough I prefer camel-casing, I think that is more susceptible to change. For people lke me who like camel-case, they can always enable the IDE option and it will remember it.

SteveMcNeill:
Closing to resubmit with Capitalization issues.

Issue #7, opened 2022-04-24
Provide a proper QB64 installer, package MinGW compilers separately


mkilgore:
I think there's a few goals here (in no particular order):

1. Provide a Windows installer for QB64.
2. Provide the 32-bit and 64-bit MinGW compilers in separate installers - they can be optional components of the QB64 installer, and downloaded during the install of QB64 if wanted.
a. This has the big advantage that the initial download of QB64 can be quite small, and then the compilers are only grabbed if necessary.
3. Remove the compilers from the repository.
4. Provide portable versions of QB64 as well that don't require installing
  a. Potentially we offer portable versions with or without the compiler. The 'without compiler' version would be useful if you already installed the compiler.
  b. This would be a lot of options, we may want to be careful to avoid creating _too_ many installation options.
  c. We would need 64-bit and 32-bit versions of all these installation options.

Some challenges I see:

- ./internal/c/c_compiler and friends are hardcoded into qb64.bas, obviously that needs to be changed.
  - The installers can easily set environment variables indicating where the compiler is, so that seems like a reasonable way to discovering them at runtime. We can check a list of locations to determine where it is.
  - A configuration option in QB64 could also be useful but I don't think QB64 currently has any runtime configuration, so that could be messy.
- The MinGW files can be moved into a separate repository, and the installers can be built from there. We will need to host them somewhere for the installer to download them. Potentially github is an acceptable place.
- QB64 itself needs changes to be realistically installable via a Windows installer. I don't think `./internal/temp` is really acceptable, because the install directory isn't generally supposed to be read/write. Ideally we move `./internal/temp` into the actual temp folder, but that's a big change. We also compile and change files outside of  `./internal/temp`, so that's not the only thing that would need addressing.
  - Potentially this can be avoided for the time being by not installing QB64 in `Program Files`, but somewhere else (maybe user specific in AppData or something) and then it can have the read/write access to its directory that it wants. This is probably a good first step considering how much of a challenge it would be to fix it otherwise.
  - These fixes are something that would benefit making packages for Linux as well, so there's reasons beyond just Windows to look into it.

a740g:
@mkilgore If we can pull this off, then I can package QB64-PE as an MSIX and get it published to the Microsoft Store. This will surely make it more discoverable. I maintain a bunch of OSS on the MS Store and they are all Win32 apps. See https://apps.microsoft.com/store/search?hl=en-in&gl=in&publisher=Samuel%20Gomes

[...]

PR #243, opened 2022-11-11
setup_win enhancements

a740g:
This is a simple modification of the setup_win setup script for Windows. Some of these are the same changes that I contributed to https://github.com/QB64Official/qb64/commit/4dc97cbca7074f65a49a2c8548d080e98947a63e

What changed?

1. We now get the latest MINGW binaries from https://github.com/niXman/mingw-builds-binaries/releases. Doing this ensures we get the latest version of GCC (`12.2.0` at the time of writing)
2. Updating to later versions of MINGW is simply a matter of changing the 32-bit and 64-bit URLs in the script
3. The script now changes the drive and path to the QB64pe directory before doing anything. Doing so allows us to call the script from anywhere
4. On 64-bit Windows, the script gives the option to download either 32-bit or 64-bit MINGW like before. However, it now chooses 64-bit automatically if there is no response after 60 seconds
5. `7zr.exe` is automatically downloaded using `curl` from the official website for unpacking the MINGW archive
6. The IDE is only launched if no additional parameters are supplied on the command line. This behavior is similar to `setup_osx` and `setup_lnx`


SteveMcNeill:
If we're going to overhaul the setup script, there's a few things that I think we should consider doing:

1) We should remove the compiler *completely* from the setup process.  Since we're getting it from the niXman's github now and using 7z to extract it, here's what I think would be a much better option for handling things:

1a) Add another dialog like we have for "Choose 32 or 64 bit versions", which should be something like, "Choose Portable (Self-Contained) or Standard (Windows User-Based) Installation."   If the portable build is chosen, then we download the compiler from the website and put it in internal/c/c_compiler/ as usual.

1b) If Standard Build is chosen, we would then check %AppData% for a QB64 folder, and see if it has a c_compiler_bit_version###.### folder in it.  (For example, for this version, it'd check for a "c_compiler32_12.2.0" folder or "c_compiler64_12.2.0" folder.)  If folder is found, then there's no reason to download and extract the c compiler -- we already have a version saved on the user's PC.  At this point, inside the QB64 c_compiler folder, we'd basically just write a single file "Redirect.txt" with contents as simple as "%AppData%/qb64pe/c_compiler64_12.2.0".   Conceivably, QB64-PE would be tweaked to simply read file, if it exists, and redirect and use that folder to find the c-compiler, rather than always assuming it's going to be at "internal/c/c_compiler/".

Once the proper compiler is downloaded and placed wherever the user wants it (either inside their AppData folder, or inside the internal/c/c_compiler folder (for a portable version such as QB64 on an USB stick), then the compiler dialog closes/moves on and we run the installation process as usual with the build of the exe happening.

************************************

The advantage I'm seeing for this approach is  a rather sizeable change in the size of the downloads which we offer for people.  If you check our latest release page, you'll see that Linux users (who have the c_compiler completely separate already) only have to download about a 9MB file.  Windows users, on the other hand, end up downloading 75-80MBs every time we do an update.

We don't actually swap out compilers all that often.  Folks shouldn't need to download and keep a half dozen copies of "compiler 12.2.0" on their machine just because they're trying to stay current with our versions and yet not wanting to get rid of  old folders for whatever reason...

We're to the point now where we're pushing out new releases every week or two.  Our user base doesn't need to have to download that whole compiler each and every time we push out an update.  If we're going to redo setup with a new script, let's redo it right, and make an extra option which will place that compiler in a permanent folder on the user's PC.   Then we can offer Standard (very small) downloads for everyone, or Portable versions (which have the compiler self-contained in internal/c/c_compiler for USB portability.)

Just my 2.5 cents worth on the issue here.  :)

mkilgore:
Sorry I've been slow :P 

> We're to the point now where we're pushing out new releases every week or two. Our user base doesn't need to have to download that whole compiler each and every time we push out an update. If we're going to redo setup with a new script, let's redo it right, and make an extra option which will place that compiler in a permanent folder on the user's PC. Then we can offer Standard (very small) downloads for everyone, or Portable versions (which have the compiler self-contained in internal/c/c_compiler for USB portability.)

I like the idea, I feel like it's a separate card though. The setup script part is relatively easy, but the build changes will get a little complicated.

Additionally I'm not sure we need to offer a script to install MinGW for them, anybody who wants to manage their own compiler copy could just install their own MinGW copy out-of-band and tell us where it is, that's probably the most straight-forward approach. If we did offer a script I think would have to be separate from the `setup_win.bat` because Window's users downloading a release don't need to compile QB64-PE, so they don't need the `make` steps and such.

mkilgore:
@a740g unfortunately we also need to change [.ci/bootstrap.bat](https://github.com/QB64-Phoenix-Edition/QB64pe/blob/main/.ci/bootstrap.bat), it downloads the MinGW compiler during CI. Perhaps the logic could be unified somehow so that `.ci/bootstrap.bat` calls `setup_win.bat`, it might get messy though...

a740g:
@SteveMcNeill I get your point. However, currently the build system expects the compiler to be in `internal/c/c_compiler`. So, like Matt said, the build system needs to be reworked. Honestly, I'd prefer the build system to detect if I have MSYS2 environment with GCC setup and use that instead. The more I think about it, the more it starts to look like https://github.com/QB64-Phoenix-Edition/QB64pe/issues/7,

a740g:
> @a740g unfortunately we also need to change [.ci/bootstrap.bat](https://github.com/QB64-Phoenix-Edition/QB64pe/blob/main/.ci/bootstrap.bat), it downloads the MinGW compiler during CI. Perhaps the logic could be unified somehow so that .ci/bootstrap.bat calls setup_win.bat, it might get messy though...

Good point. I'll see if I can merge that logic.

Now that I am thinking about it, we can have a setup script (`setup_mingw`) with the sole responsibility to download MINGW. We can then call this script from both `setup_win` and `.ci/bootstrap.bat`. That way if we ever need to update the MINGW binaries, all we need to do is modify `setup_mingw`. That should make managing the whole thing a lot easier.

SteveMcNeill:
> > @a740g 
> Now that I am thinking about it, we can have a setup script (`setup_mingw`) with the sole responsibility to download MINGW. We can then call this script from both `setup_win` and `.ci/bootstrap.bat`. That way if we ever need to update the MINGW binaries, all we need to do is modify `setup_mingw`. That should make managing the whole thing a lot easier.

That's what I was talking about above.  "We should remove the compiler completely from the setup process."

One batch file to download and extract the compiler wherever the user would like.  The second batch file to setup qb64pe.

What I'm curious about with the internals is:  Linux versions already know to use whatever compiler the user has on their system, and wherever it's located it.  How hard is it to make Windows do the same?  Look in "internal/c/compiler" first, and if not found, then look "%AppData%/Roaming/qb64pe/c_compiler", or similar. ;)

a740g:
> What I'm curious about with the internals is: Linux versions already know to use whatever compiler the user has on their system, and wherever it's located it. How hard is it to make Windows do the same? Look in "internal/c/compiler" first, and if not found, then look "%AppData%/Roaming/qb64pe/c_compiler", or similar. ;)

You can see that this stuff in https://github.com/QB64-Phoenix-Edition/QB64pe/blob/main/Makefile#L65. Well, this just means more work for @mkilgore 😉 Basically, the makefile will need have some logic to set `CC`, `CXX` and friends with paths where the binaries are at.

mkilgore:
> What I'm curious about with the internals is: Linux versions already know to use whatever compiler the user has on their system, and wherever it's located it. How hard is it to make Windows do the same? Look in "internal/c/compiler" first, and if not found, then look "%AppData%/Roaming/qb64pe/c_compiler", or similar. ;)

Well on Linux and OSX the compiler can be assumed to be in the `PATH` so we don't actually need to know where it is, the system finds it for us. The situation on Windows is a bit trickier, I don't think MinGW typically puts itself on the `PATH` even for a "normal" install (whatever that looks like), and I'm not sure it's a good idea to do so. But we can probably just have the user supply the location and/or guess a couple like Steve is suggesting.

I do think it's possible, it's just more work and I only have so much time :P If someone else wants to try it though then like @a740g said you mostly just need to change those variables in the `Makefile` so that the path can be provided via the command line. Additionally there's probably one or two special cases that need to be handled such as the [`soundfont.o`](https://github.com/QB64-Phoenix-Edition/QB64pe/blob/main/internal/c/parts/audio/extras/build.mk#L62) creation.

tests/build.mk
```diff
@@ -5,7 +5,7 @@ TEST_CFLAGS-y := -I'./tests/c/include' \
 			   -I$(PATH_LIBQB)/include \
 			   -g -std=gnu++11
 
-TEST_CFLAGS-$(win) += -mconsole
+TEST_CFLAGS-$(win) += -mconsole -static-libgcc -static-libstdc++
```
mkilgore, line 8: Thanks for fixing this :)

mkilgore (approved):

PR #245, opened 2022-11-13
Improve dialog window association and _InputBox$ support on Windows

mkilgore:
The commit messages have more detailed information, but a summary of the changes:

1. The previous handling of determining what Window handle to associate with dialogs and notifications was mostly just `GetForegroundWindow()`, which is pretty bad since it just gives the focused Window which might not even be ours. We now use three approaches (in this order) to get a window to associate with dialogs and notifications:
  a. Use `EnumWindows()` and find a top-level window associated with our Process ID.
  b. Try using `GetConsoleWindow()` and see if it gives us back a valid handle.
  c. Create a hidden window.
2. The _InputBox$ dialog was improved, it's now always on the top and tabbing through the controls works correctly.
3. I swapped the test images from the previous #241 PR to be `png` instead of `bmp`. For the tests it doesn't matter which we use, but `png`s are smaller and also show up natively on GitHub.
4. I fixed up our headers to include `stdint.h` on Windows (which we were already using but not including directly).


a740g:
> 4\. I fixed up our headers to include `stdint.h` on Windows (which we were already using but not including directly).

Thank you for doing this! Honestly, I think that all new stuff we are writing should use the `stdint.h` typedefs.

a740g (approved):

SteveMcNeill (approved):

mkilgore:
> > 4. I fixed up our headers to include `stdint.h` on Windows (which we were already using but not including directly).
> 
> Thank you for doing this! Honestly, I think that all new stuff we are writing should use the `stdint.h` typedefs.

I agree, I've pretty much been doing that already.

Issue #46, opened 2022-05-09
QB64 should offer built-in support for HTTP and HTTPS


mkilgore:
The current QB64 networking only extends to basic TCP streams and UDP. This is certainly useful, but it's not all that usable for a lot of basic tasks people actually want to do with networking, in particular HTTP and HTTPS. Writing your own HTTP handler in QB64 is fairly hard to actually accomplish, and frankly not something you should really have to do. HTTPS is also effectively impossible since it would require implementing TLS support yourself. The best option is simply pulling in external libraries via `DECLARE LIBRARY`. The QB64 IDE itself just shells out to `curl` to download Wiki pages, effectively giving up on the idea of doing it itself.

The basics of doing this is really not that hard, as `libcurl` is MIT licensed and can do all of this with a fairly simple interface. The hard part would be the interface offered, and whether it can be added into the existing networking functions (`_OPENCLIENT()` and friends) or requires some new ones.

mkilgore:
This has come up a lot. and after looking into `libcurl` I think it's decently straight-forward to get a basic version of this working. Going along with my suggestion in #21 I'll hide the initial version behind an `$UNSTABLE:HTTP` type of command while we determine what the interface should look like. I think we can make it work with the existing `_OPENCLIENT()` interface with some extensions for HTTP-specific settings, but we'll have to see. I don't think those extensions are necessary for the initial version, but if it's too hard to integrate the HTTP settings (cookies, headers, etc.) then a new API may be warranted.

mkilgore:
This is just for downloading. not rendering or anything along those lines. It's intended to replace the various existing solutions such as shelling out to `curl` or writing your own HTTP logic. Both cases have pretty large drawbacks (shelling to `curl` has a lot of overhead and is not always available, custom written HTTP logic tends to be quite fragile and cannot support HTTPS), and using HTTP is such a common use case that you really shouldn't need to come up with such a custom solution, which is the idea behind this issue.

Unfortunately I haven't haven't gotten to work on this in a while, but the backend part for `libqb` is complete. Once it's done you'll be able to just do `_OPENCLIENT("https://example.com")` to talk to HTTP endpoints without needing any special logic. Eventually we'll likely need support for handling HTTP things like supply headers, cookies, etc. but I don't think that's necessary for the initial version.

Spriggsy95:
[...]

PR #246, opened 2022-11-14
Add libcurl to the repository

mkilgore:
`libcurl` is going to be used as the backing library for implementing HTTP and HTTPS support for #46. This copy being added to the repo is used for building on Windows, for Linux and OS X we'll used the system-provided copy.

I'm doing this in its own PR to make the actual HTTP changes easier to review :) This PR just adds the `libcurl` files into `internal/c/parts/network/http/` and doesn't touch anything else - as such they're not actually used by anything yet.


a740g:
`internal/c/parts/network/src.c` and `internal/c/parts/network/log.txt` are empty files and have been living in the source tree for a long time. I wanted to check this with you but forgot. I think those serve no purpose and can be safely removed?

mkilgore:
> `internal/c/parts/network/src.c` and `internal/c/parts/network/log.txt` are empty files and have been living in the source tree for a long time. I wanted to check this with you but forgot. I think those serve no purpose and can be safely removed?

Good catch 🤦 Yeah those serve no purpose, I'll remove them in my PR to add the initial HTTP support. There might be an `#include` somewhere for that `src.c` so we might not be able to just remove it outright, but obviously it's not doing anything.

a740g (approved):

RhoSigma-QB64 (approved):

PR #252, opened 2022-11-22
Wiki dev

RhoSigma-QB64:
This changes the Help page update/download from our Wiki to use the new built-in `libcurl` support. By that the external `curl` command is no longer needed for the download.

As we had certificate problems in the past on old systems with an "Out Of Date" root CA store, the new routines will give some flexibility to the user offering some fallback options.

By default the Wiki download is using secure connections via the HTTPS protocol. However, if that fails the user is ask to go with unsecure connections via the HTTP protocol and can also decide if this is a "Do Once" decision or if the choice shall be saved permanently for the future.


source/ide/ide_methods.bas
```diff
@@ -18908,29 +18908,44 @@ FUNCTION ideupdatehelpbox
...
                 ELSE
                     f$ = CHR$(0) 'no dir scan for 'qb64pe -u' (build time update)
                 END IF
-
-                'Prepend core pages to list
-                f$ = CHR$(0) + "Keyword_Reference_-_By_usage.txt" + f$
-                f$ = CHR$(0) + "QB64_Help_Menu.txt" + f$
-                f$ = CHR$(0) + "QB64_FAQ.txt" + f$
+                et# = TIMER(0.001) - st#: IF et# < 0 THEN et# = et# + 86400
+                IF 1.25 - et# > 0 THEN _DELAY 1.25 - et#
```
mkilgore, line 18919: What purpose is the delay for? Does that avoid getting errors from `qb64phoenix.com` or something to that effect?

source/ide/wiki/wiki_methods.bas
```diff
@@ -1068,6 +1075,51 @@ FUNCTION wikiSafeName$ (page$) 'create a unique name for both case sensitive & i
...
+        IF _MESSAGEBOX("QB64-PE Help", "Can't make secure connection (https:) to Wiki, shall the IDE use unsecure (http:) instead?", "yesno", "warning" ) = 1 THEN
+            IF _MESSAGEBOX("QB64-PE Help", "Do you wanna save your choice permanently for the future?", "yesno", "question" ) = 1 THEN
+                wikiBaseAddress$ = "http://" + MID$(wikiBaseAddress$, 9)
+                WriteConfigSetting generalSettingsSection$, "WikiBaseAddress", wikiBaseAddress$
+            END IF
+            wik$ = "http://" + MID$(wik$, 9): GOTO retry
+        END IF
+    END IF
+END IF
+IF ch& = 0 THEN EXIT FUNCTION
```
mkilgore, line 1097: You can check `_StatusCode` immediately after you know `ch& <> 0`, so you could do it here rather than in the loop, it might simplify things a bit.

source/ide/wiki/wiki_methods.bas
```diff
@@ -38,22 +39,22 @@ FUNCTION Wiki$ (PageName$) 'Read cached wiki page (download, if not yet cached)
...
-            a$ = a$ + "The requested help page is not yet cached locally. To download the help page from the online Wiki, "
-        END IF
-        a$ = a$ + "a tool called ''curl'' is required, but it wasn't found on your system." + CHR$(10) + CHR$(10)
-        a$ = a$ + "* To get ''curl'', visit the official [https://curl.se/download.html download page]." + CHR$(10)
-        a$ = a$ + "** Grab the latest ''binary'' archive available for your system." + CHR$(10)
-        a$ = a$ + "** Unpack and drop the ''curl'' executable into the '''qb64pe''' folder." + CHR$(10)
-        a$ = a$ + "** If there's a file named ''curl-ca-bundle.crt'' or similar, drop it into the '''qb64pe''' folder too." + CHR$(10)
-        Wiki$ = a$: EXIT FUNCTION
-    END IF
+    'IF _SHELLHIDE("curl --version >NUL") <> 0 THEN
```
mkilgore, line 42: Personally I'm a fan of just removing code outright rather than commenting it out. It's in the `git` history so people can see the old system that way if they want to.

source/ide/wiki/wiki_methods.bas
```diff
@@ -1068,6 +1075,51 @@ FUNCTION wikiSafeName$ (page$) 'create a unique name for both case sensitive & i
...
+res$ = "": st! = TIMER
+DO
+    _DELAY 0.05
+    GET ch&, , rec$
+    IF LEN(rec$) > 0 THEN st! = TIMER
+    res$ = res$ + rec$
+    IF _STATUSCODE(ch&) = 200 THEN
+        le& = LOF(ch&)
+        IF le& > -1 THEN
+            IF LEN(res$) = le& THEN
```
mkilgore, line 1108: This works fine but a little unnecessary, checking `EOF(ch&)` always works so you could skip the `LOF(ch&)` logic and just always do the `EOF(ch&)` check to end the loop.

mkilgore (approved):
Thanks for making the changes for https vs. http 😄

PR #254, opened 2022-11-24
Bugfix to _FONT

SteveMcNeill:
Fonts 9, 15, and 17 can *only* be used in SCREEN 0.  Attempting to use them in graphic screens results in a seg fault and an instant program crash on Windows.  This is definitely the most undesirable of behaviors for a program, and can easily be caught and dealt with just by tossing a simple "Illegal Function Call" error for the issue.

### DIFF
--- a/internal/c/libqb.cpp
+++ b/internal/c/libqb.cpp
@@ -25788,16 +25788,31 @@ void sub__font(int32 f, int32 i, int32 passed) {
     i2 = 0;
     if (f == 8)
         i2 = 1;
-    if (f == 9)
+    if (f == 9) {
         i2 = 1;
+        if (!im->text) {
+            error(5);
+            return;
+        } // font 9 can *only* be used on text surfaces
+    }
     if (f == 14)
         i2 = 1;
-    if (f == 15)
+    if (f == 15) {
         i2 = 1;
+        if (!im->text) {
+            error(5);
+            return;
+        } // font 15 can *only* be used on text surfaces
+    }
     if (f == 16)
         i2 = 1;
-    if (f == 17)
+    if (f == 17) {
         i2 = 1;
+        if (!im->text) {
+            error(5);
+            return;
+        } // font 17 can *only* be used on text surfaces
+    }
     if (f >= 32 && f <= lastfont) {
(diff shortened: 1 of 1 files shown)


internal/c/libqb.cpp
```diff
@@ -25788,16 +25788,31 @@ void sub__font(int32 f, int32 i, int32 passed) {
...
         i2 = 1;
+        if (!im->text) {
+            error(5);
+            return;
+        } // font 15 can *only* be used on text surfaces
+    }
     if (f == 16)
         i2 = 1;
-    if (f == 17)
+    if (f == 17) {
```
mkilgore, line 25809: Since they all do the same thing could you roll them all into one `if (f == 9 || f == 15 || f = 17) ...`?

mkilgore (approved):

internal/c/libqb.cpp
```diff
@@ -25761,7 +25761,7 @@ int32 fontopen(qbs *f, int32 size, int32 options) { // Note: Just a stripped dow
 
 void sub__font(int32 f, int32 i, int32 passed) {
     //_FONT "?[,?]"
-    static int32 i2;
+    int32 i2 = 0; //no need for a static variable here, when we just init it to zero before ever using it.
```
mkilgore, line 25764: Agree with the change, the comment is unnecessary though :P

internal/c/libqb.cpp
```diff
@@ -25785,33 +25785,11 @@ void sub__font(int32 f, int32 i, int32 passed) {
...
-        i2 = 1;
-    if (f == 15) {
-        i2 = 1;
-        if (!im->text) {
-            error(5);
-            return;
-        } // font 15 can *only* be used on text surfaces
-    }
-    if (f == 16)
+    if ((f == 8) || (f == 9) || ((f > 13) && (f < 18))) // built-in fonts 8, 9, 13, 14, 15, 16 should be valid
```
mkilgore, line 25788: The code looks correct, I think the comment has a typo in that `13` isn't valid and `17` is.

mkilgore (approved):

Issue #257, opened 2022-11-25
When using a _StartDir$-relative path with `-x` and no `-o` the exe is placed in the wrong directory.


mkilgore:
This is a behavior issue with the changes from #201. With that change if you specify a `_StartDir$`-relative source file then `-o` is supposed to be treated as relative to `_StartDir$` as well. This works, but if you omit `-o` then the exe is placed relative to QB64-PE, which I think is bad behavior. Instead, if you specify a `_StartDir$`-relative source without specifying an output file then the executable should be placed in `_StartDir$` (with the name of the source file, with `.bas` swapped for `.exe` as we already do). This is a slight change in behavior but I think it's what everybody would expect when using the `_StartDir$`-relative source (and is how basically every compiler and command line program works, such as `gcc`, `clang`, etc.).

Issue #256, opened 2022-11-25
Use of certain commands like `$NoPrefix` screw up the compiler's command line output.


mkilgore:
Fairly straight forward, if you use some commands like `$NoPrefix` then multiple `Beginning C++ output from QB64 code...` lines appear on the command line.

It appears to be due to the `do_recompile` jump that is used to make `$NoPrefix` work, it sets some flags and then just starts the compilation process over again which then prints that message. It should be easy enough to watch for this and avoid duplicating that line.

Issue #255, opened 2022-11-24
Makefile needs to use DEP_HTTP for BUILD_QB64


mkilgore:
This is missing, which means QB64 gets built without http support. That doesn't brake the build but we'll mean the wiki download won't work.

PR #258, opened 2022-11-26
Fix QB64-PE dependency list, _StartDir$ relative compile with no -o, and command line output.

mkilgore:
This PR combines three relatively small fixes for #255, #256, and #257:

1. The `Makefile` was missing the `DEP_HTTP` flag which is now required for QB64-PE's help to work.
2. Use of some commands like `$NoPrefix` no longer causes the "Beginning C++ output..." to get duplicated.
3. When using `-x` with a `_SourceDir$`-relative source file, if no `-o` flag is provided the executable is now placed into `_SourceDir$` (instead of being placed at the location of QB64-PE).

Fixes: #255
Fixes: #256
Fixes: #257

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -175,6 +175,7 @@ endif
 	DEP_ICON := y
 	DEP_ICON_RC := y
 	DEP_SOCKETS := y
+	DEP_HTTP := y
 endif
 
 include $(PATH_INTERNAL_C)/libqb/build.mk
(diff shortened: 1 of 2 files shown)


a740g (approved):

RhoSigma-QB64 (approved):

RhoSigma-QB64:
Don't forget to pin the respective issues before merge.

mkilgore:
> Don't forget to pin the respective issues before merge.

Yeah good call, it's annoying because GitHub is smart enough to close the issues automatically due to me listing them in the commits, but not smart enough to associate the issues in the commits with the PR 🤦 

Fixes: #255
Fixes: #256
Fixes: #257

Issue #233, opened 2022-11-02
Fix `$IF` to do proper version comparisons with `VERSION`


mkilgore:
We have a `VERSION` define provided by the compiler which is intended to be used with `$IF` to optionally compile stuff based on the language version. Unfortunately it just does a string comparison between version numbers, so it doesn't work as intended. We should change `$IF` (or just comparisons to `VERSION`) such that if it sees a value resembling a 3-dot version number it compares the individual values as integers.

To see the problem, if you run this:

```vb
$IF VERSION > 3.10.0 THEN
Print 10
$END IF
```

It prints 10, but the `Print` statement should actually be skipped since the version is `3.4.0` in my case.

PR #262, opened 2022-11-26
Fix to Issue #233 with $IF and VERSION

SteveMcNeill:
https://github.com/QB64-Phoenix-Edition/QB64pe/issues/233

This should fix the comparisons with $IF and VERSION.   A few things of note here, which might affect future version labeling:

This expects to find version numbers which follow our current style -- ##.##.##   -- 3 numbers, separated by periods.

Adding in alphabetic characters will basically get ignored, and versions with less than 3 values will generate FALSE results in all comparisons.


RhoSigma-QB64 (approved):

source/qb64pe.bas
```diff
@@ -26063,6 +26077,40 @@ SUB increaseUDTArrays
     REDIM _PRESERVE udtenext(x + 1000) AS LONG
 END SUB
 
+FUNCTION CompareVersions (v1$, v2$)
+    DIM AS LONG p1, p2, p3, p4, p5, p6 'part 1 through 6
+
+    IF GetVersionValues(v1$, p1, p2, p3) = 0 THEN CompareVersions = 666: EXIT FUNCTION 'Evil Error code.  Something went wrong!
+    IF GetVersionValues(v2$, p4, p5, p6) = 0 THEN CompareVersions = 666: EXIT FUNCTION 'Evil Error code.  Something went wrong!
```
mkilgore, line 26084: I think this should produce an actual compilation error, that way people can't hit it accidentally. It's not super easy for them to test the version comparisons, so they might think that Ex. `$IF VERSION < 3.5 THEN` is working even though it's actually always treated as false.

source/qb64pe.bas
```diff
@@ -26063,6 +26077,40 @@ SUB increaseUDTArrays
...
+    IF p1 > p4 THEN CompareVersions = 1: EXIT FUNCTION
+    IF p2 < p5 THEN CompareVersions = -1: EXIT FUNCTION
+    IF p2 > p5 THEN CompareVersions = 1: EXIT FUNCTION
+    IF p3 < p6 THEN CompareVersions = -1: EXIT FUNCTION
+    IF p3 > p6 THEN CompareVersions = 1: EXIT FUNCTION
+
+    CompareVersions = 0 'if all else fails, the two versions are equal.
+END FUNCTION
+
+FUNCTION GetVersionValues (version$, v1 AS LONG, v2 AS LONG, v3 AS LONG)
```
mkilgore, line 26096: I think version comparisons with only two or one digit should be allowed, there's certain kinds of checks that can't easily be done otherwise. Ex. If you want to check for a minor version less than `3.4`, you can't just do `$IF VERSION < 3.4 THEN` so you'd have to basically use an arbitrarily large patch version to try and catch all the potential versions we release, such as `$IF VERSION < 3.3.999 THEN`.

source/qb64pe.bas
```diff
@@ -25610,16 +25615,27 @@ FUNCTION EvalPreIF (text$, err$)
             END IF
 
             IF INSTR(symbol$, ">") THEN 'check to see if we're greater than in any case with >
+                IF l$ = UserDefine(0, 7) THEN 'we're comparing VERSION numbers
```
mkilgore, line 25618: What if `VERSION` is on the right side?

source/qb64pe.bas
```diff
@@ -25599,7 +25599,12 @@ FUNCTION EvalPreIF (text$, err$)
             END IF
             IF INSTR(symbol$, "=") THEN 'check to see if we're equal in any case with =
                 UserFound = 0
+                IF l$ = UserDefine(0, 7) THEN 'we're comparing VERSION numbers
```
mkilgore, line 25602: I mentioned it on Discord but this would be a good candidate for some [tests](https://github.com/QB64-Phoenix-Edition/QB64pe/blob/main/docs/testing.md#qb64-based-tests). They get a little interesting since the version can't be assumed to be constant (the tests will get run with new versions), but it should still be somewhat straight forward to do some basic tests, Ex:

```vb
$LET verCopy = VERSION

$IF VERSION < 4.0.0 THEN
Print "4.0.0 major comparison: Success"
$ELSE
Print "4.0.0 major comparison: Failure"
$END IF

$IF 3.1.0 < VERSION THEN
Print "3.1.0 minor comparison: Success"
$ELSE
Print "3.1.0 minor comparison: Failure"
$END IF

$IF VERSION = verCopy THEN
Print "Equal comparison: Success"
$ELSE
Print "Equal comparison: Failure"
$END IF

' ...
```

Also test the error cases (which IMO should produce a compilation error, which we can also write tests for via an `.err` file):

```vb
' Should produce a "Cannot compare 'asdfasdf' as a version number" error
$IF asdfasdf = VERSION THEN
Print "Failure"
$END IF
```

PR #268, opened 2022-11-28
Addition of $AutoSave to the IDE

SteveMcNeill:
A way to spam and fill up users hard drives!  What's not to love about it?  :D

Actually this adds the ability for users to be able to set up an AutoSave feature with their QB64 programs that are open and active in the IDE.

For this to work, the following conditions must be fulfilled.  (There's a couple to sets involved, to prevent accidental carry-over to unwanted drives/user's.)

1) The IDE has to have [AutoSave: ON] down at the bottom left of the IDE.  One can just click it to toggle the states.
2) The program must set a timed amount for how often one wishes for the autosave to save their work.   $AUTOSAVE:15 <-- this, for instance, will configure AutoSave to run every 15 minutes and to save a back-up of the users work in "internal/autosave".

Note that I tried to build this will a little bit of intelligence built into it.  If nothing changes inside one's code (say you leave the IDE open while going out to eat, or going to bed overnight), it won't save multiple copies of the exact same code to the drive.  It checks for changes, and only IF there has been changes, and IF the elapsed time has passed, then does it autosave the file to drive.


SteveMcNeill:
Canceling this for now, until I can sit down and work up a proper interface with it which'll make everyone happy. ;)

PR #270, opened 2022-11-28
Only call GLUT functions from the GLUT thread.

### LINKED ISSUE #66: `glutIconifyWindow()`, `glutHideWindow()`, `glutShowWindow()` should not be called outside of the GLUT thread

mkilgore:
The main goal of this PR is fixing #66, which was necessary to fix to get my GLUT test suite working consistently (as without it being fixed, programs simply explode too often and fail the tests). This required the addition of some new code, which I placed into the `libqb/src`. This gave a good opportunity for moving all the GLUT initialization (which is fairly self-contained) into `libqb/src` as well, which what a lot of the changes here are about. The commits contain fairly detailed explanations of the changes so I would refer to those, but here's an overview:

To fix #66 we needed a way of telling the GLUT thread to perform GLUT actions when the program wants to do them, which I implemented in `glut-message.h`, `glut-message.cpp`, and `glut-msg-queue.cpp`. It's fairly simple, we have a `std::queue` protected with a mutex, and the GLUT idle or timer function (depending on platform) calls a function to process the queue. Additionally there are `libqb_glut_*` functions which can be called in the same way as the original `glut` functions and simply add commands to the queue. For most commands we don't wait for them to be completed, but for `glutGet()` we use a `completion` to wait for the command to be executed and then get the result.

The rest of the changes are refactoring the GLUT stuff into `libqb/src`. For the most part it's just a copy-paste into new functions and then wiring those back up to `libqb.cpp`. `$Console:Only` vs. GLUT have different cpp files for linking (replacing the `#ifdef` logic in `libqb.cpp`).

Additionally I was very happy that this refactor removed the little bit of Objective-C++ that was living in `libqb.cpp`, so I put that in a `.mm` file in `libqb/src/` and was able to remove the `libqb.mm` file.

Fixes: #66


SteveMcNeill:
Will this also allow us some better control over the DPI Awareness settings?  People keep asking about them, and if we can enable/disable them.  The last time I was playing around with them, they seemed to be interwoven with the same glut processes that you're moving around and tweaking here for us.

mkilgore:
@SteveMcNeill I haven't looked into it that closely but I wouldn't expect these changes to make it any easier. Really I didn't change a whole lot here, we call things slightly differently to avoid the threading issues from #66 but all the GLUT logic is the same.

a740g (approved):

RhoSigma-QB64 (approved):

PR #274, opened 2022-12-18
Expanded keyhit/keydown reporting

SteveMcNeill:
Tweaked keyhit for control, alt, shift reporting keydown events for keycodes 16, 17, 18.

Tweaked the BLEEP out of keydown.  It now reports keydown codes for CTRL +: 
ESC
`1234567890-=
Backspace
~!@#$%^&*()_+
Tab

Along with CONTROL, SHIFT, ALT codes 16, 17, 18.

That's 32 CTRL + keys which weren't reporting their downstate for us!

If there's any that I've missed, kindly report them to me and I'll tweak them into the fix as well.  For a first pass through the keyboard, I found a lot more missing codes than what I was expecting to begin with!

### DIFF
--- a/internal/c/libqb.cpp
+++ b/internal/c/libqb.cpp
@@ -32232,24 +32232,26 @@ int32 func__keyhit() {
         if (x == 100303 || x == 100304) {
             if (flag == 0) {
                 flag = 1;
-                return 16;
+                return 100016;
             }
             flag = 0;
         }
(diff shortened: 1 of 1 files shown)


SteveMcNeill:
Closing this until I see what/how Matt is going to fix the OpenGL side of things.  I can always revisit this patch, if his changes don't pan out in the future.  ;)

PR #275, opened 2022-12-18
Reduce the number of times nm is invoked

flukiluke:
Currently nm (tool to extract symbol names from libraries) is run for every relevant Declare Library function. For large libraries from which many functions are used, this can significantly impact compile times.

This change only runs nm once per library, per compile. Output is stored in a per-library file in internal/temp, which is then read as needed. From my measuring in Windows (in a VM) with a large library, this reduces the time spent searching for each function from 34 seconds to 5.5 seconds.

The 5 seconds is spent reading in the nm output. It would be nice to only read this data in once per compile, but that would require a slightly more significant code change to keep it in memory.

I have tested this on Windows and Linux but not MacOS. Relevant compile_tests pass.

### DIFF
--- a/source/qb64pe.bas
+++ b/source/qb64pe.bas
@@ -12577,7 +12577,7 @@ END IF
 'Clear nm output from previous runs
 FOR x = 1 TO ResolveStaticFunctions
     IF LEN(ResolveStaticFunction_File(x)) THEN
-        s$ = tmpdir$ + "nm_output_" + StrReplace$(StrReplace$(ResolveStaticFunction_File(x), pathsep$, "."), ":", ".") + ".txt"
+        s$ = MakeNMOutputFilename$(ResolveStaticFunction_File(x))
         IF _FILEEXISTS(s$) THEN KILL s$
     END IF
 NEXT x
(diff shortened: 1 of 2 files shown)


source/qb64pe.bas
```diff
@@ -12574,19 +12574,30 @@ IF GenerateLicenseFile AND NOT NoExeSaved THEN
     makeline$ = makeline$ + " GENERATE_LICENSE=y"
 END IF
 
+'Clear nm output from previous runs
+FOR x = 1 TO ResolveStaticFunctions
+    IF LEN(ResolveStaticFunction_File(x)) THEN
+        s$ = tmpdir$ + "nm_output_" + StrReplace$(StrReplace$(ResolveStaticFunction_File(x), pathsep$, "."), ":", ".") + ".txt"
```
mkilgore, line 12580: It would be nice to make the filename conversion here into a separate function, then the logic wouldn't need to be repeated a few times down below. You could probably put it in `utilities/build.bas` as `CreateNMOutputFile$()` or something like that.

source/qb64pe.bas
```diff
@@ -12574,19 +12574,30 @@ IF GenerateLicenseFile AND NOT NoExeSaved THEN
...
 
     'resolve static function definitions and add to global.txt
     FOR x = 1 TO ResolveStaticFunctions
+        nm_output_file$ = tmpdir$ + "nm_output_" + StrReplace$(StrReplace$(ResolveStaticFunction_File(x), pathsep$, "."), ":", ".") + ".txt"
         IF LEN(ResolveStaticFunction_File(x)) THEN
 
             n = 0
-            SHELL _HIDE "cmd /c internal\c\c_compiler\bin\nm.exe " + CHR$(34) + ResolveStaticFunction_File(x) + CHR$(34) + " --demangle -g >internal\temp\nm_output.txt"
+            IF NOT _FILEEXISTS(nm_output_file$) THEN
+                SHELL _HIDE "cmd /c internal\c\c_compiler\bin\nm.exe " + CHR$(34) + ResolveStaticFunction_File(x) + CHR$(34) + " --demangle -g >" + CHR$(34) + nm_output_file$ + CHR$(34)
```
mkilgore, line 12596: I made an `AddQuotes$()` function which just adds the `CHR$(34)`s around a string, it might simplify these lines a bit if you use it.

mkilgore (approved):
Good to see you! :)

Good find, it's hard to believe we were spending so much time on this step 🤦 I had naively assumed that basically all the time was spent on C++ compilation or source conversion, but clearly that's not the case. Perhaps we should add some level of profiling to the build that lists out how long the various parts of it took, that might reveal some other spots like this one where relatively small changes could have large improvements. I'll make a card for that :)

a740g (approved):

RhoSigma-QB64:
>The 5 seconds is spent reading in the nm output. It would be nice to only read this data in once per compile, but that would require a slightly more significant code change to keep it in memory.

Once this PR is merged I'll go in and add this using the simplebuffer system already used for code buffering, is pretty much a simple replace of the `OPEN` and `LINE INPUT` lines, once I've added the `FileToBuf` function into the system.

RhoSigma-QB64 (approved):

SteveMcNeill (approved):

Issue #232, opened 2022-11-01
Sound playback from embedded data


KingMocker:
Create a `_SNDOPENBUFFER` type of command that just takes a string that already contains the file contents for the sound you want to play. 
So, you could store a regular mp3 (or similar) in your exe and play it directly.

Issue created as requested by DSMan195276 after a post in a thread the Help Me forum.

a740g:
Just to be clear... So, is this like opening a sound file from memory rather than reading it out of storage? Honestly, I like this, and I would like something like this for image files too. I've seen plenty of code where people embed stuff in the executable and then write it to storage and then load it using `_SNDOPEN` or `_LOADIMAGE`. Although it works, it is slow (due to performance penalty of hitting slow storage) and sometimes may not be possible in environments where there is no writeable storage, or the storage has run out of space.

My idea would be to simply extend _SNDOPEN and _LOADIMAGE to load stuff directly from memory.

For sounds (soundHandle& = _SNDOPEN(fileName$[, capabilities$])) we could do:
```VB
handle& = _SNDOPEN(mybuffer$, "memory") ' here "memory" would mean a memory buffer instead of a file name
```

For images (handle& = _LOADIMAGE(filename$[, mode%])) we could do:
```VB
handle& = _LOADIMAGE(mybuffer$, 32 OR 1024) ' here 1024 would set a bit for a memory buffer instead of a file name
```

Also linking https://github.com/QB64-Phoenix-Edition/QB64pe/issues/28 to this.

RhoSigma-QB64:
Best idea ever :)

a740g:
> Best idea ever :) But mayby calling it `_LOADIMAGEMEM` and `_SNDOPENMEM` respectively instead of overloading the existing command names.

But this would mean we would add two new keywords and must write up two new wiki pages. I already wrote eight and my fingers hurt.😝 And then there will be some people who'll call us out and say we are adding bloat. 😉

Just kidding. I am good with this too.

SpriggsySpriggs:
Very cool idea. Would make embedding images and sound files so much easier. No more BIN2BAS files. Ooooh. I'd also be able to put that into effect for my FLAC player and my API stuff that grabs images from the web. Very exciting stuff, guys.

SpriggsySpriggs:
Maybe you can add a switch to those keywords. Something like `_FromString`. So `_LoadImage _FromString string$`

Issue #217, opened 2022-10-24
Issues with the PLAY statement


barcoboy:
According to the QB64 help text, the PLAY statement "N" parameter can be used to play a note "between 0 to 84 in the 7 octaves, where 0 is a rest).  However, the 0 value to N does not play a rest, but plays a low C note instead.   The following code should play 4 beeps with rests in between, but instead plays a high G, then a low C, high G, low C, etc.

PLAY "T120O4L4GN0GN0GN0GN0"

Actually all "N" notes should be shifted up by one as compared to QuickBASIC 4.5.  Right now, N0 in QB64 is N1 in QB4,5, which is a low C.  An octave higher N12 in QB64 is N13 in QB4.5, and so on.

Also the sound of the notes in QB4.5 is a louder square waveform as compared to the softer triangle waveform of QB64... wondering if that could be changed as well?

a740g:
> Also the sound of the notes in QB4.5 is a louder square waveform as compared to the softer triangle waveform of QB64... wondering if that could be changed as well?

I actually got square waveform working in my [miniaudio-fixes ](https://github.com/a740g/QB64pe/tree/miniaudio-fixes) branch but removed it. It was much simpler and shorter than the current implementation and seemed to work quite well except for cases where `PLAY` played multiple notes together (see example 5 in https://qb64phoenix.com/qb64wiki/index.php/PLAY). Well, mixing two square waves just sounds awful (or probaly we can do better). I'll investigate this in detail later. But for now, we'll stick to the current implementation.

The other fixes here are done and ready for review. Thanks to @mkilgore for figuring out the octave fix.

Issue #186, opened 2022-09-21
Optimize `GenerateWaveform`


a740g:
`GenerateWaveform` in `audio.cpp` generates stereo waveform which is used by `SOUND` and `PLAY`. `SOUND` and `PLAY` have no stereo capabilities and hence generating stereo waveform is a waste of memory and processing resources.

`GenerateWaveform` should be rewritten to generate mono waveform instead.

barcoboy:
As I mentioned in #217, the sound generated by PLAY is a triangle waveform, where QB45 generates a near square waveform.

Issue #167, opened 2022-09-06
Fix timing of `SOUND` command.


mkilgore:
Currently the blocking aspect of `SOUND` is handled by a `Sleep()` call [here](https://github.com/QB64-Phoenix-Edition/QB64pe/blob/main/internal/c/parts/audio/audio.cpp#L758). The problem with doing it this way is that it's not extremely accurate, the calculated time is shortened compared to the actual length to attempt to account for overhead that happens after the sound is started. The current result is a bit too fast, and `SOUND` returns before it finishes playing (but still slower than older QB64 version, which are even more off of the correct timing).

We can likely achieve accurate timing of the sound by simply using the `completion` API rather than a `Sleep()`. We can trigger a `completion` when the sound is done, which would immediately unblock the QB64 thread at the right time. Implementing this may require doing #166 first, but potentially not as long as we can somehow hook it into a callback from `miniaudio` signalling us that the sound is done.

PR #279, opened 2022-12-30
Miniaudio fixes

a740g:
This PR fixes a bunch of stuff and adds a few.

#28 - Adds a new function `_SNDNEW`
```vb
soundHandle& = _SNDNEW(frames&, channels&, bits&)
```
This allows one to create an empty sound buffer in memory. The buffer memory can then later be acquired using `_MEMSOUND` and read from and written to using `_MEMGET` & `_MEMPUT`.

#166 - `_SNDRAW` and friends are now fully implemented using miniaudio data source APIs. Sample data is automatically pulled from the miniaudio thread and synchronization is handled using the libqb mutex APIs. This also fixes a PLAY(n) function bug that @SteveMcNeill found.

#167 - Because of the above change, we can now use exact sample times (without compensations). However, we are still relying on `Sleep()`. Seems to work much better than before in my tests.

#186 - `GenerateWaveform()` is now optimized to generate mono waveform instead of stereo. As such, all functions (`SOUND`, `PLAY`) that rely on it have been updated as well.

#217 - `PLAY` `N` commands now work correctly. Thanks to @mkilgore for figuring out the octave fix. `N0` now correctly rests. I've left a note in the issue explaining why I abandoned work on square waveform for `SOUND` and `PLAY`.

#232 - This feature request has been partially implemented. Sounds can now be opened from memory. We can close this issue once image and font memory loading has been implemented.

#280 - Added Amiga AHX & HVL decoder.

Added tests for `_SNDNEW` and `_SNDOPEN`.

Miniaudio has been updated to version 0.11.11.


SteveMcNeill:
Just my 2 cents here, but you might want to make the command _SNDNEW instead of _NEWSOUND.  It'd fit the existing syntax for all the rest of the sound commands, and would fall right into the same area for the wiki and visual reference and all.

a740g:
> Just my 2 cents here, but you might want to make the command _SNDNEW instead of _NEWSOUND. It'd fit the existing syntax for all the rest of the sound commands, and would fall right into the same area for the wiki and visual reference and all.

I actually thought about it and ended up naming it like `_NEWIMAGE`. Now that you mentioned this, I think `_NEWSND` might be more fitting.

RhoSigma-QB64 (approved):

SteveMcNeill (approved):

internal/c/parts/audio/audio.cpp
```diff
@@ -703,61 +622,47 @@ static ma_uint8 *GenerateWaveform(double frequency, double length, double volume
...
     if (!data)
         return;
 
+    auto samples = bytes / SAMPLE_FRAME_SIZE(float, 1);
+
     // Move data into sndraw handle
-    for (i = 0; i < bytes; i += SAMPLE_FRAME_SIZE(ma_int16, 2)) {
-        audioEngine.soundHandles[audioEngine.sndInternal]->rawQueue->PushSampleFrame((float)((ma_int16 *)(data + i))[0] / 32768.0f,
-                                                                                     (float)((ma_int16 *)(data + i))[1] / 32768.0f);
+    for (auto i = 0; i < samples; i++) {
```
mkilgore, line 656: If this turns out to be a bit slow we could add a `PushSampleFrames()` function which takes the full buffer and length, then we would only need to lock the mutex once for the whole copy, rather than once per sample. Since this is on the producer side performance is less of a concern though.

internal/c/parts/audio/extras/build.mk
```diff
@@ -34,14 +34,22 @@ LIBXMP_OBJS += $(patsubst %.c,$(PATH_INTERNAL_C)/parts/audio/extras/libxmp-lite/
 LIBXMP_LIB := $(PATH_INTERNAL_C)/parts/audio/extras/libxmp-lite.a
 
 $(PATH_INTERNAL_C)/parts/audio/extras/libxmp-lite/%.o: $(PATH_INTERNAL_C)/parts/audio/extras/libxmp-lite/%.c
-	$(CC) $(CFLAGS) -Wall -DLIBXMP_CORE_PLAYER -DLIBXMP_NO_PROWIZARD -DLIBXMP_NO_DEPACKERS -DBUILDING_STATIC $< -c -o $@
+	$(CC) -O2 $(CFLAGS) -Wall -DLIBXMP_CORE_PLAYER -DLIBXMP_NO_PROWIZARD -DLIBXMP_NO_DEPACKERS -DBUILDING_STATIC $< -c -o $@
 
 $(LIBXMP_LIB): $(LIBXMP_OBJS)
 	$(AR) rcs $@ $^
 
+HIVELY_SRCS := hvl_replay.c
```
mkilgore, line 42: It would probably be worth making a new issue for the addition of AHX and HVL support. You can keep the code in this PR, but having the issue will make it easier for people searching for this change.

internal/c/parts/audio/extras/build.mk
```diff
@@ -34,14 +34,22 @@ LIBXMP_OBJS += $(patsubst %.c,$(PATH_INTERNAL_C)/parts/audio/extras/libxmp-lite/
...
+
+HIVELY_OBJS += $(patsubst %.c,$(PATH_INTERNAL_C)/parts/audio/extras/hivelytracker/%.o,$(HIVELY_SRCS))
```
mkilgore, line 44: We need the license for this guy :)

internal/c/parts/audio/miniaudio_impl.cpp
```diff
@@ -12,51 +12,39 @@
 //
 //-----------------------------------------------------------------------------------------------------
 
-//-----------------------------------------------------------------------------------------------------
-// HEADER FILES
-//-----------------------------------------------------------------------------------------------------
 // Enable Ogg Vorbis decoding
 #define STB_VORBIS_HEADER_ONLY
 #include "extras/stb_vorbis.c"
```
mkilgore, line 17: Nice to see the Pulseaudio issues resolved :)

tests/compile_tests/audio_mem_test/sndopen_mem_test.bas
```diff
@@ -0,0 +1,1311 @@
+$Console:Only
+Option _Explicit
+Option _ExplicitArray
+
+Dim buffer As String: buffer = LoadTestData
+Print "Size ="; Len(buffer)
+
+
+Dim h As Long: h = _SndOpen(buffer, "memory")
```
mkilgore, line 9: Thank you for adding this test, I was wondering how this worked :) This is pretty handy, it's probably just as good as being able to pass a `_MEM` in.

tests/compile_tests/audio_mem_test/newsound_test.bas
```diff
@@ -0,0 +1,14 @@
+$Console:Only
```
mkilgore, line 1: If you feel like it, a few error tests wouldn't hurt. Ex. Passing negatives, or zero.

mkilgore (approved):

internal/c/parts/audio/audio.cpp
```diff
@@ -93,337 +71,286 @@ extern ptrszint dblock;         // Required for Play(). Did not find this declar
...
-    /// This initializes the queue and calculates the sample frames per block
-    /// </summary>
-    /// <param name="pmaEngine">A pointer to a miniaudio engine object</param>
-    /// <param name="pmaSound">A pointer to a miniaudio sound object</param>
-    SampleFrameBlockQueue(ma_engine *pmaEngine, ma_sound *pmaSound) {
-        first = last = nullptr;
-        blockCount = frameCount = maEngineTime = sampleFramesPlaying = bufferUpdatePosition = 0;
+/// @brief A miniaudiio raw audio stream datasource
+struct RawStream {
+    ma_data_source_base maDataSource;         // miniaudio data source (this must be the first member of our struct)
```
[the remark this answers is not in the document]
a740g: I was worried about that too. But since we are not using memset on the struct objects we should be good. I checked the code on the miniaudio side too and it does not do anything funny. I compiled and tested this on VC++ and it worked there as well.

internal/c/parts/audio/audio.cpp
a740g: Totally agree. I'll add that just in case. However, we cannot use `PushSampleFrames()` for `sub__sndraw` since it only pushes one frame at a time.

internal/c/parts/audio/audio.cpp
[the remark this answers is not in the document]
mkilgore: Well I was thinking that because there's usually a VTable pointer placed at the beginning of a class/struct in C++ the pointer to `ma_data_source_base` wouldn't be directly convertible to a `RawStream`. But if it wasn't working I would expect at least the `RawStreamOnRead()` callback to explode, so I'm probably missing something, Maybe this class/struct doesn't have a VTable? I don't know the details on when C++ would add one, you don't have any `virtual` functions so maybe that's why.

We have tests that play actual sounds so I think we would notice if it was exploding :D

internal/c/parts/audio/extras/build.mk
a740g: Done.

internal/c/parts/audio/audio.cpp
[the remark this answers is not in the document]
a740g: Yes. A vtable will be added when the class has virtual functions.

tests/compile_tests/audio_mem_test/newsound_test.bas
a740g: Done.

internal/c/parts/audio/extras/build.mk
a740g: I knew I was missing something. 😊
Getting this added.

a740g: Done.

SteveMcNeill (approved):

source/subs_functions/subs_functions.bas
```diff
@@ -2058,6 +2058,19 @@ id.arg = MKL$(ULONGTYPE - ISPOINTER)
...
 regid
 
+' a740g: Feature request #28
+clearid
+id.n = qb64prefix$ + "SndNew" ' Name in CaMeL case
+id.Dependency = DEPENDENCY_AUDIO_OUT ' QB64-PE library dependency
+id.subfunc = 1 ' 1 = function, 2 = sub
+id.callname = "func__sndnew" ' C/C++ function name
+id.args = 3 ' number of arguments "passed"
+id.arg = MKL$(LONGTYPE - ISPOINTER) + MKL$(LONGTYPE - ISPOINTER) + MKL$(LONGTYPE - ISPOINTER) + MKL$(LONGTYPE - ISPOINTER) ' arguments & types
```
mkilgore, line 2068: Did you intend for this to list 4 arguments instead of 3?

a740g: That's a good catch. I initially wanted 4 args. The 4th one being frequency. But I dropped it. Now I cannot remember why.  
I wanted frames to be mandatory with the rest being optional.
But yeah. This is clearly an issue. I forgot to change `id.arg`.

PR #286, opened 2023-01-09
Image load from memory support

a740g:
This PR implements loading image files from memory. See #232 for more details.

The arrangement is like what we have for the recent `_SNDOPEN(sound$, "memory")` addition.

`_LOADIMAGE` is now updated with the following syntax:
- `handle& = _LOADIMAGE(fileName$[, [mode&][, requirements$]])`
- `requirements$` can be one or more of the following:
  - `hardware` - should be loaded as a hardware image
  - `adaptive` - should be loaded with an adaptive palette
  - `memory` - should be loaded from memory instead of a file

We will leave #232 open until we finish implementing loading fonts from memory.


mkilgore:
The only thing I'm kinda hesitant on is using a bit flag to indicate loading from memory vs. file, it feels a bit too obscure. `_SNDOPEN()` is better in this regard since it at least takes a string list, so it was a bit more natural and also easier to read. I think we should consider having either a separate function for it, or maybe an extra parameter instead of the bit flag 🤷

a740g:
> The only thing I'm kinda hesitant on is using a bit flag to indicate loading from memory vs. file, it feels a bit too obscure. `_SNDOPEN()` is better in this regard since it at least takes a string list, so it was a bit more natural and also easier to read. I think we should consider having either a separate function for it, or maybe an extra parameter instead of the bit flag 🤷

I agree with your point. I'd prefer using `_LOADIMAGE` rather than making a separate function just to keep things consistent with the sound stuff. Maybe we can change `_LOADIMAGE` as the following:
`handle& = _LOADIMAGE(image$, [bpp], [capabilities$])`

Now this opens another can of worms or probably I got all this wrong. Can we have more than one optional parameter for functions? From what I have seen it does not work that way.

a740g:
> So more than a single optional param isn't hard to deal with. It's just looking and grabbing those flags from the right places so they match. ;)

Yeah. That's what I thought too. Expect that it did not work that way for me for `Functions` on the *BASIC side*. It does work correctly for `Subs`.
Probably I was screwing up `id.specialformat`. Or `id.specialformat` has different format for `Functions`? I am not sure.

Do you have any documentation on `id.specialformat`? I'll dig through the sources when I get some time.

a740g:
Implemented an overloaded version of `_LOADIMAGE`.

source/subs_functions/subs_functions.bas
```diff
@@ -1063,7 +1063,7 @@ id.args = 2
 id.arg = MKL$(STRINGTYPE - ISPOINTER) + MKL$(LONGTYPE - ISPOINTER)
 id.specialformat = "?[,?]"
 id.ret = LONGTYPE - ISPOINTER
-id.hr_syntax = "_LOADIMAGE(fileName$[, mode])"
+id.hr_syntax = "_LOADIMAGE(fileName$[, mode& | mode$])"
```
mkilgore, line 1066: Not to throw a huge wrench in things, but I had thought you'd just add a third argument to go along with bpp, rather than turn the second one into a string. TBH I don't know which is better, but having an optional third argument would probably be a fair amount simpler from an implementation standpoint 🤷

RhoSigma-QB64:
The most important thing is not to break the existing behavior. `_LOADIMAGE(file$, 32)` must still remain working, If people would need to change that to `_LOADIMAGE(file$, "32")` in their existing programs, then it would be very annoying. But looking on your `hr_syntax` above I guess you did take care of that.

RhoSigma-QB64 (approved):

source/subs_functions/subs_functions.bas
a740g: Ya, we need both the second and the third argument to be optional.  That is what I originally intended to do. And honestly, that would have been far easier and simpler had it worked the way I expected it to.

Just to give a little background...
While I was working on the dialog box stuff, I figured that you cannot have more than one optional argument for functions. The `passed` argument works like a `bool` in the case of functions. It works like a bit field for Subs. Also, if you look at [subs_functions.bas](https://github.com/QB64-Phoenix-Edition/QB64pe/blob/main/source/subs_functions/subs_functions.bas), we do not have a single function that takes more than one optional argument. The only exception to that is `_SCREENIMAGE` but that makes all argument optional in one go and `func__screenimage` handles `passed` as a `bool`.

Let me give it a go again on the different branch. Probably, I'll be able to explain it better then.

mkilgore: Well that is unfortunate, I did not realize that :-/ I wonder how hard it would be to fix that, it's probably not pretty though...

mkilgore (approved):

PR #288, opened 2023-01-10
Add missing words to syntax highlighter

SteveMcNeill:
_ScaledWidth and _ScaledHeight are two keywords which never got added to the syntax highlighter.  This corrects that.


a740g:
There is nothing about these in the wiki too. We need to add those as well.

a740g (approved):

Issue #294, opened 2023-01-15
On Windows, using Sleep in Console programs will not trigger any timers.


mkilgore:
Pretty simple, Windows has custom code in `sub_sleep` for console programs which checks for console input to end the sleep. The logic doesn't call `evnt()` at regular intervals like the other sleep implementations (limit, delay, sleep not on console) which means things like timers will fail to trigger during a sleep. Overall it's pretty simple to fix.

Additionally the console sleep does not check `sleep_break` or `stop_program`. This makes some sense given it doesn't trigger timers anyway, but after it is fixed to trigger timers it should also check these flags (which will do things such as ending the sleep if a timer goes off).

Issue #293, opened 2023-01-15
If a timer expires while stopped it does not get triggered immediately when turned on.


mkilgore:
Pretty simple, if you stop a timer with `TIMER STOP` and the timer expires while it's stopped, then it is supposed to trigger immediately when `TIMER ON` is done, but that doesn't happen. Instead it will trigger a random amount of time later.

The issue is that `TIMER ON` does not set the `qbevent` flag to check for events after it. When the timer expires it will set `qbevent`, but that happens when the timer is stopped and thus the timer doesn't actually trigger. It then won't trigger even when turned on until events are checked again.

Issue #273, opened 2022-12-10
ON TIMER() not firing on 1st interval


grymmjack:
It seems that `ON TIMER()` is not firing on the 1st interval, if it is the first timer set with `_FREETIMER`
I am using `QB64-PE Compiler V3.4.1`

### What I expect to happen:
`TIMER_A_TICKS%` gets incremented every 5 seconds.
`TIMER_B_TICKS%` gets incremented every second.

### What is happening:
`TIMER_A_TICKS%` gets incremented on the 10th second, then every 5 seconds.
`TIMER_B_TICKS%` gets incremented every second.

```basic
DIM SHARED TIMER_A AS INTEGER
DIM SHARED TIMER_A_TICKS AS INTEGER
DIM SHARED TIMER_B AS INTEGER
DIM SHARED TIMER_B_TICKS AS INTEGER

TIMER_A% = _FREETIMER
TIMER_B% = _FREETIMER

TIMER_A_TICKS% = 1
TIMER_B_TICKS% = 1

ON TIMER(TIMER_A%, 5) timer_a_update
ON TIMER(TIMER_B%, 1) timer_b_update
TIMER(TIMER_A%) ON
TIMER(TIMER_B%) ON

DO:
    _LIMIT 30
    LOCATE 1, 1 : PRINT _TRIM$(STR$(TIMER_A_TICKS%));
    LOCATE 3, 1 : PRINT _TRIM$(STR$(TIMER_B_TICKS%));
    _DISPLAY
LOOP UNTIL _KEYHIT = 27
SYSTEM

SUB timer_a_update
    TIMER_A_TICKS% = TIMER_A_TICKS% + 1
END SUB

SUB timer_b_update
    TIMER_B_TICKS% = TIMER_B_TICKS% + 1
END Sub
```

https://user-images.githubusercontent.com/412263/206874388-593ccc93-9030-46d6-981a-01e9c60ed6c8.mp4

Thanks

gs-rickchristy:
Here is an old post that Vince found on the old forum that seems related:
https://qb64forum.alephc.xyz/index.php?topic=3835.msg131830

grymmjack:
Thank you @mkilgore for this fix!

PR #295, opened 2023-01-15
Fix timers at program start, TIMER STOP, and Sleep with console

mkilgore:
Three timer-related fixes in here:

1. Timers started at the start of the program take two intervals to trigger. This was due to the faulty `last_time == 0` logic, which was done after check if the interval had passed. This doesn't work correctly at program start when `time_now == 0`. It's not necessary to check the interval anyway (we're just going to reset it), so we can just check for the `last_time == 0` special case first and avoid the issue.
2. If a timer expired during `TIMER STOP`, it wouldn't trigger when `TIMER ON` is used. This was simple to fix, `TIMER ON` wasn't setting `qbevent = 1` to tell the program to check for events (like timer expiration). This meant the timer wouldn't trigger until something else set `qbevent`.
3. Using `Sleep` with a console program on Windows won't run timers while sleeping. This was due to the custom sleep logic not calling `evnt()`. I rewrote it to sleep for short intervals and then call `evnt()` to handle timers and such, like all the other sleep commands do.

Fixes: #273 
Fixes: #294 
Fixes: #293


mkilgore:
This is ready to be reviewed :)

a740g (approved):

Issue #303, opened 2023-02-06
Lack of support for multiple optional Function arguments


a740g:
See the conversation in #286

QB64-PE supports optional `Sub` arguments, but it does not support more than one optional `Function` argument. This limits the ability to create flexible and user-friendly functions.

**For example**, if we were to extend the built-in keywords and add a statement like MessageBox title$, message$, style$, all three arguments could be made optional. However, if we write a built-in function like MessageBox&(title$, message$, style$), only one of these could be implemented as optional.

The proposed solution is to enhance the language to allow for multiple optional `Function` arguments, bringing it in line with other modern dialects of BASIC.

This would provide a significant improvement in the functionality and usability of the language and would allow for the creation of more versatile functions.

PR #305, opened 2023-02-07
Add support for optional function arguments

mkilgore:
Currently functions only have very limited optional argument support, this expands it so that we can have more complex sets of optional arguments for functions, such as multiple arguments where not all need to be provided. This will be used in the future for some upcoming functionality.

Note that this does not support any generic optional argument format, the commas always have to be provided unless an optional argument is at the end of the parameter list. Thus, if you have a format with one required and two optional arguments and you want to omit the first optional one, then you need to call it as `foo(2, , 3)`, rather than `foo(2, 3)`. This is important for avoiding ambiguous situations, and is how many SUBs currently handle this.

The two functions that violate that requirement are INSTR() and _INSTRREV(), which use the format `[?],?,?` and allow omitting the comma for the first argument. This format is simply handled as a special case.

Fixes: #303

---

This doesn't add any functions that make use of the new optional parameters, but I did test the likely new parameters for `_LOADIMAGE()` which is `?[,[?][.?]]` and they work as expected, allowing any combination of the second and third parameter to be provided but requiring the first.

### DIFF
--- a/source/qb64pe.bas
+++ b/source/qb64pe.bas
@@ -16423,27 +16423,36 @@ FUNCTION evaluatefunc$ (a2$, args AS LONG, typ AS LONG)
 
     SetDependency id2.Dependency
 
+    argCount = countFunctionElements(a$)
+    ReDim providedArgs(argCount)
+
     passomit = 0
-    omitarg_first = 0: omitarg_last = 0
+    hasOptionalFirstArg = 0
+    firstOptionalArgument = 0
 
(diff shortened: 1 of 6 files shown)


a740g (approved):
Thank you for the tests. Those make it so much easier to understand. I hope one day we'll be able to break down `qb64pe.bas` into more manageable and logical parts.

RhoSigma-QB64 (approved):

Issue #301, opened 2023-01-26
QB64-PE fails to find symbols in a `.so` library if it only has a dynamic symbol table.


mkilgore:
When using `Declare Library` with a `.so` library, QB64-PE will attempt to use `nm` to view the symbol table of the `.so` file to find the symbol being called by a `Sub`/`Function` within the `Declare Library` section. The real purpose here is to determine whether the given symbol name is a `C` symbol or a `C++` symbol, as how you go about calling the symbol changes depending on that detail.

The issue here is that a `.so` file has two symbol tables in it, the "regular" symbol table that typically includes all symbols present at compile time, and the "dynamic" symbol table that only includes the symbols necessary for linking and loading the `.so`. To find a symbol it is necessary to check both tables (and if not, checking the dynamic one would be the better of the two).

We do attempt to check both tables, you can tell `nm` to display the dynamic table instead of the regular symbol table via passing the `-D` flag, which we do. Unfortunately, this was broken by #275, we now only have one `nm` output file per `.so` that we scan and never run `nm` again with the `-D` flag. Specifically, the second call to `nm` now has a `IF NOT _FILEEXISTS(nm_output_file$) THEN` check around it that it didn't used to have, which means it never runs since the first `nm` file has always been generated by that point via the other `nm` call without the `-D` flag.

Note that I'm not entirely sure if such libraries were working in `v3.4.1` and below as I found them to simply give different errors when attempting to compile. When fixing `v3.5.0` to properly have two separate output files for both `nm` commands they started working.

The automated testing we have didn't catch this because I didn't realize there were two symbol tables at the time I made those tests. As a result, `liblib.so` still has its regular symbol table and we don't have a test of a `.so` file that only has a dynamic symbol table. It's easy enough to create one by running `strip` on the existing `liblib.so`, and when doing that the test starts failing. We should keep the existing test and add a new one for the stripped `.so`.

PR #307, opened 2023-02-07
Fix DECLARE LIBRARY against stripped .so file

mkilgore:
.so files can be stripped such that they contain no "regular" symbol table but do still contain the "dynamic" symbol table, this is pretty typical for .so files. QB64-PE is supposed to check both tables when linking against a .so file, but a bug in ab0c2b18 meant that the second run of nm with the -D flag to check the dynamic symbol table no longer happens. The fix is to introduce a new output file for the dynamic run so that they are handled separately in terms of caching the result.

A new test .so file that only contains a dynamic symbol table was added to avoid this in the future.

Fixes: #301


source/qb64pe.bas
```diff
@@ -12577,7 +12577,10 @@ END IF
 'Clear nm output from previous runs
 FOR x = 1 TO ResolveStaticFunctions
     IF LEN(ResolveStaticFunction_File(x)) THEN
-        s$ = MakeNMOutputFilename$(ResolveStaticFunction_File(x))
+        s$ = MakeNMOutputFilename$(ResolveStaticFunction_File(x), 0)
```
RhoSigma-QB64, line 12580: Not sure if we should do a `ClearBuffers` here too, just in case the file was already loaded into memory. It can definitly not make things worst, but maybe better :), as it ignore calls with for unknown buffers (see utilities\s-buffer\sb_qb64pe_extension.bm).
i.e. instead of `IF _FILEEXISTS(s$) THEN KILL s$` write `IF _FILEEXISTS(s$) THEN KILL s$: ClearBuffers s$`

RhoSigma-QB64 (commented):
Added some notes...

source/qb64pe.bas
```diff
@@ -12577,7 +12577,10 @@ END IF
...
+        IF _FILEEXISTS(s$) THEN KILL s$
+
+        s$ = MakeNMOutputFilename$(ResolveStaticFunction_File(x), 1)
         IF _FILEEXISTS(s$) THEN KILL s$
```
RhoSigma-QB64, line 12584: Not sure if we should do a `ClearBuffers` in the two _FILEEXISTS THEN KILL lines too, just in case the file was already loaded into memory. It would definitly not break things, as it ignores calls for unknown buffers (see utilities\s-buffer\sb_qb64pe_extension.bm).
i.e. instead of `IF _FILEEXISTS(s$) THEN KILL s$` write `IF _FILEEXISTS(s$) THEN KILL s$: ClearBuffers s$`

RhoSigma-QB64 (approved):

SteveMcNeill (approved):

source/qb64pe.bas
mkilgore: Thanks for the suggestion, yeah I'll add that :)

PR #308, opened 2023-02-12
Wiki dev

RhoSigma-QB64:
Finally I got the HTML Entity and UTF-8 handling working to a sufficient grade to properly parse our Wiki pages. However, some breaking changes were required to get along with it. Some pages such as `ASCII`, `_PRINTSTRING`, `_SCREENPRINT` and others contain UTF-8 chars for the lower 32 control chars, which got problematic especially for the `Carriage Return` CHR$(13) code, which was used as line delimiter in the Help_Txt$ variable and so never was printed but created a new line which screwed up the page layout.
To be able to really display all 256 chars, another method for line break detection was needed. Fortunately the Help_Txt$ variable is an interleaved string of char/color/link values, so to be able to use the full range for chars, the line end marker had to go either into the color or link part. As the links are not really predictable (unless counting them in a pre-pass) you cannot set a defined value you can call "safe" as a constant marker, so color is the part remaining.
The color value is actually encoded like the result of the `SCREEN` function when used to get the color, i.e. the lower nibble (bits 0-3) are the foreground (0-15) and the first 3 bits of the upper nibble (bits 4-6) are the background (0-7) while the remaining bit 7 is the "blinking" foreground flag. As we don't use blinking colors in the IDE help, this bit is the perfect candidate for use as line end marker. Hence, where formerly the char position value was tested for = 13 or <> 13, you'll now see that the color position value is checked for > 127 or < 128 respectively.

The second big change targets the handling of links, templates and centered text. Formerly the text in those layout elements was simply copied over into the Help_Txt$, bypassing the remaining parser code and so never running through the Entity/UTF-8 substitution stage. To change this, the flow logic was changed so that the text is not simply copied without further processing, but instead only flags are set to indicate the processed layout element and then the main parser continues, running all chars through the substitution stage and accumulating them in the respective variables at the end.

Apart from that two big changes, some more fixes were made, mainly to filter/skip useless line feeds in the Wiki text and also some more old code was removed, which formerly handled things which I've eliminated from the Wiki pages in the meantime.
The list of recognized UTF-8 sequences has grown by 20+ more chars and the IDE will show dynamically created warnings at the top of help pages, if unknown Entities/UTF-8 are detected in it and asking people to report it in the Forum, so we can update and complete the list over time.

As a result of all that, and of the fact that I've implemented local links handling a while back already, I've now also removed the filters for all those "Return to Table of Contents" links and the "Letter rows" at the top of the alphabetic index page. All those links are now properly rendered and working just like in the online Wiki, so you can happily jump around on the pages. The IDE help pages are now probably as close as we can come to the real Wiki feeling.

### DIFF
--- a/source/ide/ide_methods.bas
+++ b/source/ide/ide_methods.bas
@@ -2376,7 +2376,7 @@ FUNCTION ide2 (ignore)
                         x = l
                         x3 = 1
                         c = ASC(Help_Txt$, x)
-                        DO UNTIL c = 13
+                        DO UNTIL ASC(Help_Txt$, x + 1) > 127
                             IF Help_Select = 2 THEN
                                 IF y >= Help_SelY1 AND y <= Help_SelY2 THEN
                                     IF x3 >= Help_SelX1 AND x3 <= Help_SelX2 THEN
@@ -2464,7 +2464,7 @@ FUNCTION ide2 (ignore)
                         x = l
                         a$ = ""
                         c = ASC(Help_Txt$, x)
-                        DO UNTIL c = 13
+                        DO UNTIL ASC(Help_Txt$, x + 1) > 127
                             lnk = CVI(MID$(Help_Txt$, x + 2, 2))
                             IF lnk THEN a$ = a$ + CHR$(c) ELSE a$ = a$ + CHR$(0) 'only add text with links
                             x = x + 4: c = ASC(Help_Txt$, x)
@@ -2485,7 +2485,7 @@ FUNCTION ide2 (ignore)
                             c = ASC(Help_Txt$, x)
                             oldlnk = 0
                             lnkx1 = 0: lnkx2 = 0
-                            DO UNTIL c = 13
+                            DO UNTIL ASC(Help_Txt$, x + 1) > 127
                                 lnk = CVI(MID$(Help_Txt$, x + 2, 2))
                                 IF lnkx1 = 0 AND lnk <> 0 AND oldlnk = 0 AND px = x2 THEN lnkx1 = x2
                                 IF lnkx1 <> 0 AND lnk = 0 AND lnkx2 = 0 THEN lnkx2 = x2 - 1
@@ -2597,7 +2597,7 @@ FUNCTION ide2 (ignore)
                 x = l
                 x2 = 1
                 c = ASC(Help_Txt$, x)
-                DO UNTIL c = 13
+                DO UNTIL ASC(Help_Txt$, x + 1) > 127
 
                     IF x2 = Help_cx THEN
                         lnk = CVI(MID$(Help_Txt$, x + 2, 2))
@@ -17723,7 +17723,7 @@ SUB Help_ShowText
             sx = Help_wx1
             c = ASC(Help_Txt$, x): col = ASC(Help_Txt$, x + 1)
             LOCATE sy, sx
-            DO UNTIL c = 13
+            DO UNTIL col > 127
                 COLOR col AND 15, col \ 16
                 IF IdeSystem = 3 AND Help_Select = 2 THEN
                     IF y >= Help_SelY1 AND y <= Help_SelY2 THEN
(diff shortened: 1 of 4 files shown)


a740g (approved):

source/ide/ide_methods.bas
```diff
@@ -17744,8 +17734,7 @@ SUB Help_ShowText
             Help_LineLen(y - Help_sy) = x3 - 1
 
             FOR x4 = 1 TO Help_wx2 - POS(0) + 1
-                IF col = 0 THEN col = 7
-                COLOR col AND 15, col \ 16
+                COLOR 7, (col - 128) \ 16
```
mkilgore, line 17737: Do you not want to set the foreground color with `col AND 15`?

source/ide/wiki/wiki_global.bas
```diff
@@ -62,42 +62,46 @@ DIM SHARED wpEntRepl(0 TO 10) AS wikiEntityReplace
...
+wpUtfReplCnt = wpUtfReplCnt + 1: wpUtfRepl(wpUtfReplCnt).utf8 = MKI$(&HA7C3): wpUtfRepl(wpUtfReplCnt).repl = CHR$(135) 'accent (�)
+wpUtfReplCnt = wpUtfReplCnt + 1: wpUtfRepl(wpUtfReplCnt).utf8 = MKI$(&HAAC3): wpUtfRepl(wpUtfReplCnt).repl = CHR$(136) 'accent (�)
+wpUtfReplCnt = wpUtfReplCnt + 1: wpUtfRepl(wpUtfReplCnt).utf8 = MKI$(&HABC3): wpUtfRepl(wpUtfReplCnt).repl = CHR$(137) 'accent (�)
+wpUtfReplCnt = wpUtfReplCnt + 1: wpUtfRepl(wpUtfReplCnt).utf8 = MKI$(&HA8C3): wpUtfRepl(wpUtfReplCnt).repl = CHR$(138) 'accent (�)
+wpUtfReplCnt = wpUtfReplCnt + 1: wpUtfRepl(wpUtfReplCnt).utf8 = MKI$(&HAFC3): wpUtfRepl(wpUtfReplCnt).repl = CHR$(139) 'accent (�)
+wpUtfReplCnt = wpUtfReplCnt + 1: wpUtfRepl(wpUtfReplCnt).utf8 = MKI$(&HAEC3): wpUtfRepl(wpUtfReplCnt).repl = CHR$(140) 'accent (�)
+wpUtfReplCnt = wpUtfReplCnt + 1: wpUtfRepl(wpUtfReplCnt).utf8 = MKI$(&HA2C2): wpUtfRepl(wpUtfReplCnt).repl = CHR$(155) 'cents
+wpUtfReplCnt = wpUtfReplCnt + 1: wpUtfRepl(wpUtfReplCnt).utf8 = MKI$(&HBDC2): wpUtfRepl(wpUtfReplCnt).repl = CHR$(171) 'fraction (�)
+wpUtfReplCnt = wpUtfReplCnt + 1: wpUtfRepl(wpUtfReplCnt).utf8 = MKI$(&HBCC2): wpUtfRepl(wpUtfReplCnt).repl = CHR$(172) 'fraction (�)
```
mkilgore, line 99: The characters in the comments got switched, not sure if you care or not 🤷

mkilgore (approved):

source/ide/ide_methods.bas
RhoSigma-QB64: Not in that place. It wouldn't make a visible difference, but would need more processing time. That loop basically just fills the empty space from help line end to window width with the current bg color, the fg color is completely irrelevant here.

source/ide/wiki/wiki_global.bas
RhoSigma-QB64: Yep, they are correct now when loaded into the IDE with CP437. The former was CP1252 characters.

PR #322, opened 2023-03-23
miniaudio version number bump

a740g:
Related to #320 

David bumped up the version number to v0.11.13 😁

### DIFF
--- a/internal/c/parts/audio/miniaudio.h
+++ b/internal/c/parts/audio/miniaudio.h
@@ -1,6 +1,6 @@
 /*
 Audio playback and capture library. Choice of public domain or MIT-0. See license statements at the end of this file.
-miniaudio - v0.11.12 - 2023-03-19
+miniaudio - v0.11.13 - 2023-03-23
 
 David Reid - mackron@gmail.com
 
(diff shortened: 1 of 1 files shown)


mkilgore:
This is in relation to #320, but this change sounds like it would let us get more accurate `SOUND` timing (that we were trying to do a while back). It could be worth looking into at some point:

> Add support for callback notifications when a sound reaches the end.

mkilgore (approved):

a740g:
> This is in relation to #320, but this change sounds like it would let us get more accurate `SOUND` timing (that we were trying to do a while back). It could be worth looking into at some point:

Yeah. This was one of the first things that caught my eye when I looked at the change logs. My initial thought was to completely get rid of the snd_mainloop(). However, after I read the documentation, I realized this may need a little bit of work.

The doc says:
```text
Use `ma_sound_at_end()` to determine whether or not a sound is currently at the end. For a looping
sound this should never return true. Alternatively, you can configure a callback that will be fired
when the sound reaches the end. Note that the callback is fired from the audio thread which means
you cannot be uninitializing sound from the callback. To set the callback you can use
`ma_sound_set_end_callback()`. Alternatively, if you're using `ma_sound_init_ex()`, you can pass it
into the config like so:

    soundConfig.endCallback = my_end_callback;
    soundConfig.pEndCallbackUserData = pMyEndCallbackUserData;

The end callback is declared like so:

    void my_end_callback(void* pUserData, ma_sound* pSound)
    {
        ...
    }
```

Like you said, we can use this to get `SOUND` working a lot better. I'll add that to my TODO list. 👍

Issue #324, opened 2023-03-24
Update compile_tests.sh to allow testing C/C++ header libraries with source / $INCLUDE relative paths


a740g:
We added the ability to use `C/C++` header libraries using paths relative to the source file or the `$INCLUDE` file. Currently `compile_tests.sh` always runs `qb64pe` executable from the same directory as the `.bas` source file, so this new logic can't ever be triggered. Hence, our tests will not go through the new code path.

We will need to update `compile_tests.sh` to allow testing header libraries with source / `$INCLUDE` relative paths (i.e. run `qb64pe` executable from outside the source directory).

See #321.

PR #326, opened 2023-03-27
Add tests for file resolution behavior, fix when compiling from command line

mkilgore:
Includes small fixes to #321 and #325 to allow them to work when compiling from the command line (That is, when `NoIDEMode` is true, then `path.source$` contains the path to the source file, not `idepath$`). Additionally it adds a bunch of tests for the file resolution behavior.

Fixes: #324


tests/compile_tests/midi/relative_to_include.bas
```diff
@@ -0,0 +1,11 @@
+$CONSOLE:ONLY
+OPTION _EXPLICIT
+
+'$Include:'include/soundfont.bi'
+
+DIM h AS LONG: h = _SNDOPEN("2SKIPIX.mid")
```
a740g, line 6: If we include the MIDI file and run and get the output, we can also test if the SF3 stuff is working or not.

a740g (approved):

tests/compile_tests/midi/relative_to_include.bas
mkilgore: Yeah good point, I should have just left it and added `.output` files :D

RhoSigma-QB64 (approved):

SteveMcNeill (approved):

Issue #188, opened 2022-09-21
Implement `falcon.h` functionality / fixes into QB64-PE


a740g:
This was already on the QB64 roadmap. See https://github.com/QB64Team/qb64/issues/72

The current QB64-PE font rasterizer(?) has issues.

![image](https://user-images.githubusercontent.com/47574584/191394127-c709412a-fed7-40d7-8bc9-c385117160e4.png)

Quoting @SteveMcNeill 
> The first 3 lines are QB64's PRINT, the next 3 are using falcon.h to print the lines.   As you can see, the top of the letters are getting cut off without enough proper spacing.

See https://qb64forum.alephc.xyz/index.php?topic=2248.msg114890#msg114890 for @RhoSigma-QB64's library based on `falcon.h`.
Also see https://github.com/FellippeHeitor/InForm/blob/development/falcon.h

We should take this opportunity to evaluate replacing [FreeTypeAmalgam](https://github.com/vinniefalco/FreeTypeAmalgam) with [stb_truetype](https://github.com/nothings/stb).

[FreeTypeAmalgam](https://github.com/vinniefalco/FreeTypeAmalgam) (2.4.12) is 10+ years old and based on an old version of [FreeType](https://github.com/freetype/freetype) and has not been updated since. The current [FreeType](https://github.com/freetype/freetype) version is 2.12.1.
`FreeTypeAmalgam.c` is a mammoth 5.59 MB file since it is a combination of all core FreeType source code.

In contrast, the current [stb_truetype](https://github.com/nothings/stb) is 5077 lines (4488 sloc) and 194 KB. It is not only tiny but has all essential features that is required by QB64-PE.

Also see https://github.com/QB64Team/qb64/issues/71

PR #330, opened 2023-04-04
Tidy up font code

a740g:
This PR is an effort (the first of many) to reach the end goal in #188.

The following changes were made:

- Consolidate required files into `internal/c/parts/video/font/` and remove unnecessary files
- Avoid including `src.c` from `libqb.cpp` and instead move font code to `font.cpp` that is compiled using a makefile
- If font support is not needed the makefile will compile a stub instead
- Update to FreeType 2.4.12 Amalgamation

In subsequent font PRs, I'll try to move more font functions to `font.cpp` until we have something that is more-or-less self-contained.

The ultimate step would be to swap out `FreeTypeAmalgam` with `stb_truetype`.


Makefile
```diff
@@ -265,12 +265,12 @@ else
 endif
 
 ifneq ($(filter y,$(DEP_FONT)),)
-	EXE_LIBS += $(QB_FONT_LIB)
-	CXXFLAGS += -DDEPENDENCY_LOADFONT
+	EXE_LIBS += $(FREETYPE_OBJS) $(FONT_OBJS)
 	QBLIB_NAME := $(addsuffix 1,$(QBLIB_NAME))
```
mkilgore, line 269: Since the define is gone we can get rid of the QBLIB_NAME addition as well :) Very good stuff, it puts us one step closer to not having any more conditional compilation for `libqb.cpp`.

internal/c/parts/video/font/font.cpp
```diff
@@ -1,350 +1,339 @@
...
+
+//[ ]what if no glyph index could be found and it returns 0??
+//[ ]FT_Done_Face
+//[ ]Check for memory leaks
+
+#include "font.h"
+#include "freetypeamalgam.h"
+#include "rounding.h"
+
+extern uint16_t codepage437_to_unicode16[];
```
mkilgore, line 14: I think this `extern` should go in `font.h`. We also might consider moving the actual declaration into this file.

internal/c/parts/video/font/build.mk
```diff
@@ -0,0 +1,22 @@
...
+	stub_font.cpp
+
+FREETYPE_OBJS := $(patsubst %.c,$(PATH_INTERNAL_C)/parts/video/font/%.o,$(FREETYPE_SRCS))
+
+FONT_OBJS := $(patsubst %.cpp,$(PATH_INTERNAL_C)/parts/video/font/%.o,$(FONT_SRCS))
+
+FONT_STUB_OBJS := $(patsubst %.cpp,$(PATH_INTERNAL_C)/parts/video/font/%.o,$(FONT_STUB_SRCS))
+
+$(PATH_INTERNAL_C)/parts/video/font/%.o: $(PATH_INTERNAL_C)/parts/video/font/%.c
+	$(CC) -O2 $(CFLAGS) -DDEPENDENCY_CONSOLE_ONLY -Wall $< -c -o $@
```
mkilgore, line 17: For these lines you can get rid of the `-DDEPENDENCY_CONSOLE_ONLY` because none of these files depend on `libqb.h` (which was causing the problems).

mkilgore (approved):

Makefile
a740g: Noted.

internal/c/parts/video/font/font.cpp
a740g: That's a good idea.👍 I'll make this change in the next PR.

PR #332, opened 2023-04-19
Fixes a potential segfault in PLAY()

a740g:
This fixes a potential segfault in `PLAY()` if it is used before `SOUND` or `PLAY`.
This also adds a tiny bit of silence after `BEEP` so that consecutive `BEEP`s sound unique and not like a long continuous sound.

### DIFF
--- a/internal/c/parts/audio/audio.cpp
+++ b/internal/c/parts/audio/audio.cpp
@@ -806,25 +806,18 @@ void sub_sound(double frequency, double lengthInClockTicks) {
     SendWaveformToQueue(data, soundwave_bytes, !audioEngine.musicBackground);
 }
 
-/// <summary>
-/// This generates a default 'beep' sound.
-/// </summary>
-void sub_beep() { sub_sound(900, 5); }
+/// @brief This generates a default 'beep' sound
+void sub_beep() {
+    sub_sound(900, 4.5);
+    sub_sound(32767, 0.5); // we'll send a very short silence after the beep so that two successive beeps sound unique
+}
 
-/// <summary>
-/// This was designed to returned the number of notes in the background music queue.
-/// However, here we'll just return the number of sample frame remaining to play when Play(), Sound() or Beep() are used.
-/// This allows programs like the following to compile and work.
-///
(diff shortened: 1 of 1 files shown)


RhoSigma-QB64 (approved):

internal/c/parts/audio/audio.cpp
```diff
@@ -806,25 +806,18 @@ void sub_sound(double frequency, double lengthInClockTicks) {
...
-/// </summary>
-/// <param name="ignore">Well, it's ignored</param>
-/// <returns>Returns the number of sample frames left to play for Play(), Sound() & Beep()</returns>
+/// @brief This was designed to returned the number of notes in the background music queue.
+/// However, here we'll just return the number of sample frame remaining to play when Play(), Sound() or Beep() are used
+/// @param ignore Well, it's ignored
+/// @return Returns the number of sample frames left to play for Play(), Sound() & Beep()
 int32_t func_play(int32_t ignore) {
-    if (audioEngine.isInitialized && audioEngine.sndInternal == 0) {
+    if (audioEngine.isInitialized && audioEngine.sndInternal == 0 && audioEngine.soundHandles[audioEngine.sndInternal]->rawStream) {
```
mkilgore, line 820: How will this eventually get initiated? I'm just thinking that if your program only has PLAY statements in it then they might never work since the rawStream will just always be null.

a740g: The initialization of `rawStream` only happens on demand the first time when `SOUND` or `PLAY (statement)` are used.

Here:
https://github.com/QB64-Phoenix-Edition/QB64pe/blob/58fab19c8a5525342d715df1e8f0da541ab3c241/internal/c/parts/audio/audio.cpp#L859

mkilgore: Ah, that's what I get for not reading close enough :D you changed the function, not the sub that actually plays the music.

mkilgore (approved):

internal/c/parts/audio/audio.cpp
a740g: Well, now that I am looking at the PR description again, I think I should have made it clearer. It is a little confusing. 😄

PR #335, opened 2023-04-28
QB64-PE font support overhaul

### LINKED ISSUE #188: Implement `falcon.h` functionality / fixes into QB64-PE

### LINKED ISSUE #232: Sound playback from embedded data

a740g:
This PR accomplishes several things:

1. Reimplements True Type font rending using a lazy glyph caching system. This improves True Type rendering performance by 6x+. The glyphs are loaded on demand
2. Implements loading font files from memory using the new `memory` `_LOADFONT` string argument
3. Implements four functions (`_UFontHeight`, `_UPrintWidth`, `_ULineSpacing`, `_UPrintString`) that includes much of the functionality of and improves on _falcon.h_

Notes:
We cannot move to `stb_truetype.h` yet because it does not support native monochrome rendering that QB64 depends on for True Type rendering in legacy SCREEN modes.
We are not moving to the full FreeType library yet (although the migration implementation is complete) due to a strange rendering alignment bug that I am currently investigating.
We can finally close #232 now that sound, image, and font functions fully support loading assets from memory.

Fixes #232
Fixes #188


RhoSigma-QB64 (approved):

internal/c/parts/video/font/font.cpp
```diff
@@ -1,339 +1,1417 @@
...
-        if (fonts[i].monospace) {
-            if (w2 != fonts[i].monospace_width) {
-                w2 = fonts[i].monospace_width;
-                ox = 0;                // no repositioning possible
-                ox2 = w2 / 2 - w1 / 2; // align to centre
+        /// @brief Creates a glyph belonging to a codepoint, caches its bitmap + info and adds it to the hash map
+        /// @param codepoint A valid UTF-32 codepoint
+        /// @return True if successful or if the glyph is already in the map
+        bool CacheGlyph(FT_ULong codepoint) {
+            if (glyphs.count(codepoint) == 0) {
```
mkilgore, line 444: God forbid they name it `.exists()` so I don't have to look up what `.count()` is supposed to do here 🤣

internal/c/libqb.cpp
```diff
@@ -37698,11 +37531,11 @@ void display() {
                         static int32 render_option;
                         static int32 ok;
                         static uint8 *rt_data;
-                        static int32 rt_w, rt_h, rt_pre_x, rt_post_x;
+                        static int32 rt_w, rt_h;
                         render_option = 1;
                         if (rt_data_last)
                             free(rt_data_last);
-                        ok = FontRenderTextUTF32(font[f], &chr_utf32, 1, render_option, &rt_data, &rt_w, &rt_h, &rt_pre_x, &rt_post_x);
+                        ok = FontRenderTextUTF32(font[f], &chr_utf32, 1, render_option, &rt_data, &rt_w, &rt_h);
```
mkilgore, line 37538: I think this is a problem, it happens on the GLUT thread rather than the QB64 thread. Since this will access/modify the `Font` object and `Glyph` cache they need to be thread-safe.

Alternatively we could potentially separate things a bit so that the GLUT thread here gets a _copy_ of the selected `Font`, I'm not really sure which option is easier.

internal/c/parts/video/font/font.cpp
```diff
@@ -1,339 +1,1417 @@
...
+            if (!newHandle)
+                return -1; // we cannot return 0 here since 0 is a valid internal handle
+
+            fonts.push_back(newHandle);
+            size_t newVectorSize = fonts.size();
+
+            // If newVectorSize == vectorSize then push_back() failed
+            if (newVectorSize <= vectorSize) {
+                delete newHandle;
+                return -1; // we cannot return 0 here since 0 is a valid internal handle
```
mkilgore, line 598: Could you use `INVALID_FONT_HANDLE` for these cases?

internal/c/parts/video/font/font.cpp
```diff
@@ -1,339 +1,1417 @@
...
+
+    FONT_DEBUG_PRINT("Allocated (%llu x %llu) buffer", strPixSize.x, strPixSize.y);
+
+    FT_Pos penX = 0;
+
+    if (fnt->monospaceWidth) {
+        for (FT_ULong i = 0; i < codepoints; i++) {
+            auto cp = codepoint[i];
+
+            if (fnt->CacheGlyph(cp)) {
```
mkilgore, line 932: It's minor, but rather than expose `CacheGlyph()` and `glyphs` directly I'd just add a `GetGlyph(cp)` that returns the `Glyph` pointer and automatically loads the cache if it's not present. It makes the caching aspect transparent to callers.

internal/c/parts/video/font/font.cpp
```diff
@@ -1,339 +1,1417 @@
...
+    // Check max width
+    if (passed & 1) {
+        if (max_width < 1)
+            return;
+    } else {
+        max_width = 0;
+    }
+
+    // Check UTF argument
+    if (passed & 2) {
```
mkilgore, line 1152: It seems like a lot of this logic is common among a few of the commands and could be moved into a separate function.

source/subs_functions/subs_functions.bas
```diff
@@ -1325,11 +1325,11 @@ clearid
 id.n = qb64prefix$ + "LoadFont": id.Dependency = DEPENDENCY_LOADFONT
 id.subfunc = 1
 id.callname = "func__loadfont"
-id.args = 3
-id.arg = MKL$(STRINGTYPE - ISPOINTER) + MKL$(DOUBLETYPE - ISPOINTER) + MKL$(STRINGTYPE - ISPOINTER)
-id.specialformat = "?,?[,?]"
+id.args = 4
+id.arg = MKL$(STRINGTYPE - ISPOINTER) + MKL$(LONGTYPE - ISPOINTER) + MKL$(STRINGTYPE - ISPOINTER) + MKL$(LONGTYPE - ISPOINTER)
+id.specialformat = "?,?[,[?][,[?]]]"
```
mkilgore, line 1330: Genuinely not sure it matters but I think it should be `[,[?][,?]]` so that you can't omit the last parameter but include the comma. In practice I don't remember how my code works that well, I'm not sure it would let you do that anyway or if it actually makes a distinction between these syntaxes.

source/subs_functions/subs_functions.bas
```diff
@@ -1325,11 +1325,11 @@ clearid
...
 id.callname = "func__loadfont"
-id.args = 3
-id.arg = MKL$(STRINGTYPE - ISPOINTER) + MKL$(DOUBLETYPE - ISPOINTER) + MKL$(STRINGTYPE - ISPOINTER)
-id.specialformat = "?,?[,?]"
+id.args = 4
+id.arg = MKL$(STRINGTYPE - ISPOINTER) + MKL$(LONGTYPE - ISPOINTER) + MKL$(STRINGTYPE - ISPOINTER) + MKL$(LONGTYPE - ISPOINTER)
+id.specialformat = "?,?[,[?][,[?]]]"
 id.ret = LONGTYPE - ISPOINTER
-id.hr_syntax = "_LOADFONT(fileName$, size%[, " + CHR$(34) + "{MONOSPACE|, BOLD|, ITALIC|, UNDERLINE|, UNICODE|, DONTBLEND}" + CHR$(34) + "])"
+id.hr_syntax = "_LOADFONT(fileName$, size&[, " + CHR$(34) + "{MONOSPACE|, UNICODE|, DONTBLEND|, MEMORY}" + CHR$(34) + "][, fontIndex])"
```
mkilgore, line 1332: I think this should match the `specialformat` in terms of where the brackets are. So, a pair of brackets around the string parameter, and the first `[,` should be closed out after the last parameter, not before it.

source/subs_functions/subs_functions.bas
```diff
@@ -1427,6 +1427,55 @@ id.ret = LONGTYPE - ISPOINTER
...
+regid
+
+clearid
+id.n = qb64prefix$ + "UPrintWidth"
+id.Dependency = DEPENDENCY_LOADFONT
+id.subfunc = 1
+id.callname = "func__UPrintWidth"
+id.args = 3
+id.arg = MKL$(STRINGTYPE - ISPOINTER) + MKL$(LONGTYPE - ISPOINTER) + MKL$(LONGTYPE - ISPOINTER)
+id.specialformat = "?[,[?][,[?]]]"
```
mkilgore, line 1450: Same thing here, the last parameter probably shouldn't be optional, and the `hr_syntax` should match.

source/subs_functions/subs_functions.bas
```diff
@@ -1427,6 +1427,55 @@ id.ret = LONGTYPE - ISPOINTER
...
+regid
+clearid
+id.n = qb64prefix$ + "UPrintString"
+id.Dependency = DEPENDENCY_LOADFONT
+id.subfunc = 2
+id.callname = "sub__UPrintString"
+id.args = 6
+id.arg = MKL$(LONGTYPE - ISPOINTER) + MKL$(LONGTYPE - ISPOINTER) + MKL$(STRINGTYPE - ISPOINTER) + MKL$(LONGTYPE - ISPOINTER) + MKL$(LONGTYPE - ISPOINTER) + MKL$(LONGTYPE - ISPOINTER)
+id.specialformat = "(?,?),?[,[?][,[?][,?]]]"
+id.hr_syntax = "_UPRINTSTRING (x&, y&), text$[, maxWidth&][, utfEncoding&][, fontHandle&]"
```
mkilgore, line 1476: The brackets at the end should reflect the `specialformat`.

tests/compile_tests/font/test.bas
```diff
@@ -1,8 +1,45 @@
...
 SYSTEM
+
+' Loads a whole file from disk into memory
+FUNCTION LoadFileFromDisk$ (path AS STRING)
+    IF _FILEEXISTS(path) THEN
+        DIM AS LONG fh: fh = FREEFILE
+
+        OPEN path FOR BINARY ACCESS READ AS fh
+
+        LoadFileFromDisk = INPUT$(LOF(fh), fh)
```
mkilgore, line 40: Huh, did not know you could use `INPUT$()` like that. I always do a silly `SPACE$()` + `GET #f` combo to read whole files 🤦

mkilgore (approved):
Good stuff, thanks for doing this :) I think the only non-trivial comment I had was around the thread-safety with the Glyph cache. If you need assistance feel free to ask but it's probably enough to just wrap the cache logic in a mutex so that only one thread is using the `unordered_map` at a time. If that creates a performance issue for some reason we could look at other options.

internal/c/parts/video/font/build.mk
```diff
@@ -14,9 +14,9 @@ FONT_OBJS := $(patsubst %.cpp,$(PATH_INTERNAL_C)/parts/video/font/%.o,$(FONT_SRC
 FONT_STUB_OBJS := $(patsubst %.cpp,$(PATH_INTERNAL_C)/parts/video/font/%.o,$(FONT_STUB_SRCS))
 
 $(PATH_INTERNAL_C)/parts/video/font/%.o: $(PATH_INTERNAL_C)/parts/video/font/%.c
-	$(CC) -O2 $(CFLAGS) -DDEPENDENCY_CONSOLE_ONLY -Wall $< -c -o $@
+	$(CC) -O2 $(CFLAGS) -Wall $< -c -o $@
 
 $(PATH_INTERNAL_C)/parts/video/font/%.o: $(PATH_INTERNAL_C)/parts/video/font/%.cpp
-	$(CXX) -O2 $(CXXFLAGS) -DDEPENDENCY_CONSOLE_ONLY -Wall $< -c -o $@
+	$(CXX) -O2 $(CXXFLAGS) -Wall $< -c -o $@
```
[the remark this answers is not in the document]
a740g: I think I removed it after we talked about it here: https://github.com/QB64-Phoenix-Edition/QB64pe/pull/330/files#diff-867e77c7e2165c2aa16dd4867dbc5e099223269e21a037c8dc2efa7c9560a55b

And then later I started using libqb.h and forgot to add it back. 🤦

internal/c/libqb.cpp
a740g: Thank you for pointing this out. I'll start woking on adding the thread-safety code in the Glyph caching system. Will update here once I am done.

internal/c/parts/video/font/font.cpp
```diff
@@ -1,339 +1,1417 @@
...
+    /// @return True if the buffer was allocated correctly
+    bool ResizeCodepointBuffer(FT_ULong count) {
+        if (count <= codepoints || !count) {
+            codepoints = count;
+            return codepoint != nullptr;
         }
-        if (FT_Load_Glyph(fonts[i].handle, glyph_index, FT_LOAD_DEFAULT)) {
-            // failed!
+
+        auto *tempBuffer = (FT_ULong *)realloc(codepoint, count * sizeof(FT_ULong));
```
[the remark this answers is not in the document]
a740g: 😂 You are right. Now why did I not think about that? I guess, this is what happens when I reuse my old code without thinking about it too much. I'll re-implement this using std::vector. Should be fairy simple.

internal/c/parts/video/font/font.cpp
```diff
@@ -1,339 +1,1417 @@
...
+    /// OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+    /// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR
+    /// IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+    /// @param state The current state of the decoder
+    /// @param codep The decoded codepoint after state changes to UTF8DecodeState::ACCEPT
+    /// @param byte The next UTF-8 byte in the input stream
+    /// @return UTF8DecodeState::ACCEPT if enough bytes have been read for a character,
+    /// UTF8DecodeState::REJECT if the byte is not allowed to occur at its position,
+    /// and some other positive value if more bytes have to be read
+    uint32_t DecodeUTF8(uint32_t *state, FT_ULong *codep, uint32_t byte) {
```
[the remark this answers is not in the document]
a740g: Yeah. I was just trying to make as minimal changes to Bjoern's original code as possible. I think he did it that way for performance reasons. Not sure. 🤷‍♂️

source/subs_functions/subs_functions.bas
SteveMcNeill: I don't think it matters.  Usually you can use a comma without having to include the optional parameter after it.

_PutImage , ,     '<-- This works just fine, for example, even though it's completely blank and those commas pass nothing

source/subs_functions/subs_functions.bas
SteveMcNeill: Just curious: What happened to the Bold, Underline, Italic here?  From what I remember, they never worked anyway; have we just removed them completely?  How will that affect existing programs that someone might have that references them?  Even though I don't currently think they do anything, I think they still compile without tossing any sort of error messages.

SteveMcNeill (approved):

source/subs_functions/subs_functions.bas
a740g: Well, they were not really doing anything. So, no point in advertising that stuff. We can still pass those, but it will be ignored just like the old code did. So, no changes in the way those were handled.

SteveMcNeill: Sounds good.   I just noticed the change and was kinda worried we'd error out with them now.  Ignoring them is the proper thing to do, IMHO, I think.  ;)

internal/c/parts/video/font/font.cpp
```diff
@@ -1,339 +1,1417 @@
...
+            1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,0,1,1,1,1,1,0,1,0,1,1,1,1,1,1, // s1..s2
+            1,2,1,1,1,1,1,2,1,2,1,1,1,1,1,1,1,1,1,1,1,1,1,2,1,1,1,1,1,1,1,1, // s3..s4
+            1,2,1,1,1,1,1,1,1,2,1,1,1,1,1,1,1,1,1,1,1,1,1,3,1,3,1,1,1,1,1,1, // s5..s6
+            1,3,1,1,1,1,1,3,1,3,1,1,1,1,1,1,1,3,1,1,1,1,1,1,1,1,1,1,1,1,1,1, // s7..s8
+        };
+        // clang-format on
+
+        uint32_t type = utf8d[byte];
+        *codep = (*state != UTF8DecoderState::ACCEPT) ? (byte & 0x3fu) | (*codep << 6) : (0xff >> type) & (byte);
+        *state = utf8d[256 + *state * 16 + type];
```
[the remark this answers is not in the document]
a740g: There is a version of this with two seperate tables on Bjoern's GitHub but I did not bother using that as it was more than 10 years old. This version of his code was being used in falcon.h. So, I decided to stick with it.

tests/compile_tests/font/test.bas
RhoSigma-QB64: It depends, AFAIK `INPUT$` if used that way, stops reading when hitting a EOT marker (CTRL-Z (ascii 26)), so `SPACE$` + GET is the safer method if you don't know the contents of a file and will be definitly the required way to go for binary files.

internal/c/parts/video/font/font.cpp
```diff
@@ -1,339 +1,1417 @@
...
 
-        if (monochrome) {
+        return 0;
+    }
+
+    /// @brief Converts an UTF-8 array to UTF-32. This does not check for BOM
+    /// @param str The UTF-8 array
+    /// @param byte_len The size of the array in bytes
+    /// @return The number of codepoints that were converted
+    FT_ULong ConvertUTF8(const FT_Bytes str, FT_ULong byte_len) {
```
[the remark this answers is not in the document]
a740g: Well, we need not use these FT types at all. I was just using those to avoid type casting and compiler warnings. But it looks like after using these, I ended up type casting more than I thought. 😁 I'll be changing all of these to `stdint` types. That'll also make it easier if we ever decide to swap out the True Type library in the future.

internal/c/parts/video/font/font.cpp
a740g: `INVALID_FONT_HANDLE` is defined as zero because that is what stuff in `libqb.cpp` expects. However, `CreateHandle()` returns 0 as a valid handle and that handle is reserved in the constructor and never passed to libqb.cpp.

internal/c/parts/video/font/font.cpp
a740g: That's a good idea.

internal/c/parts/video/font/font.cpp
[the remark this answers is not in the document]
a740g: Changed this to `uint8_t`.

internal/c/parts/video/font/font.cpp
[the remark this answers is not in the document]
a740g: Replaced most `FT_Byte`, `FT_Bytes` and `FT_ULong` with `stdint` types.

internal/c/parts/video/font/font.cpp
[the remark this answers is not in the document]
a740g: This uses `std::vector` now.

internal/c/libqb.cpp
a740g: Using mutex in most top-level functions declared in `font.h`.

internal/c/parts/video/font/font.cpp
a740g: This is done.

internal/c/parts/video/font/font.cpp
a740g: Moved these into macros.

source/subs_functions/subs_functions.bas
a740g: The last two parameters are optional. Hence the syntax. Seems to be working for me and I tried lot of permutations and combinations. 😊

source/subs_functions/subs_functions.bas
a740g: I wrote those `hr_syntax` help texts in that way to keep thing consistent with how it was done before. Like this:
https://github.com/QB64-Phoenix-Edition/QB64pe/blob/17b8fe4ac4dcd64c4d292c2191b8d6ba3dfe2fd5/source/subs_functions/subs_functions.bas#L3067

I guess the previous authors wanted to keep the help text simple for users to understand. 🤷‍♂️

source/subs_functions/subs_functions.bas
a740g: The last two parameters are optional. Same reason for `hr_syntax` as above.

source/subs_functions/subs_functions.bas
a740g: This one too due to the simplified help texts used by previous authors.

internal/c/parts/video/font/font.cpp
mkilgore: Ah, thanks, I didn't catch that it was zero :)

PR #339, opened 2023-05-01
Dialog improvements

a740g:
This PR has the following improvements:

1. It makes use of the functionality introduced by @mkilgore in #305 to make many mandatory dialog parameters optional. Optional function arguments were not available when dialog support was added to QB64-PE. Hence, this adds the changes in a non-breaking way.
2. Some dialog functions and statements take string arguments that can contain parsable options. These needed to be in lower case. This PR allows these options to be passed in any case now.
3. The MessageBox wrapper function is removed. This was placed to support InForm. However, with the latest font changes, InForm needs to be updated for QB64-PE anyway. This is something that I am already working on in a new private repo. I'll make it public once it is ready.
4. I moved much of the `qbs` stuff to `std::string` to reduce `qbs` dependency. This is something that I'll continue doing as and when I make any future changes (wherever possible).


internal/c/parts/gui/gui.cpp
```diff
@@ -76,318 +62,267 @@ static void gui_free_tokens(char **tokens) {
...
-        aIconType = qbs_new(0, 0);
+    std::string aTitle;
+    std::string aMessage;
+    std::string aIconType;
 
     if (passed & 1)
-        qbs_set(aTitle, qbs_add(qbsTitle, qbs_new_txt_len("\0", 1)));
-    else
-        qbs_set(aTitle, qbs_new_txt_len("\0", 1));
+        aTitle.assign((const char *)qbsTitle->chr, qbsTitle->len);
```
mkilgore, line 72: We might consider wrapping this up in a function just so you can't forget to pass the length. I would probably fall for that one :D Something like a `qbs_to_string(std::string *, qbs *)`. If we also had a `string_to_qbs` that produced a new `qbs *` from a `std::string` you might not even need `libqb.h` anymore, just having the `struct qbs` would be enough since you never access anything inside it.

internal/c/parts/gui/gui.cpp
```diff
@@ -433,57 +368,3 @@ bool gui_alert(const char *fmt, ...) {
...
-#    define MB_OKCANCEL 0x00000001L
-#    define MB_ABORTRETRYIGNORE 0x00000002L
-#    define MB_YESNOCANCEL 0x00000003L
-#    define MB_YESNO 0x00000004L
-#    define MB_RETRYCANCEL 0x00000005L
-#    define MB_SYSTEMMODAL 0x00001000L
-
-// This exists because InForm calls it directly.
-// It only supports the "MB_OK" and "MB_YESNO" options
-int MessageBox(int ignore, char *message, char *title, int type) {
```
mkilgore, line 456: Are we tossing this?

mkilgore (approved):
I really like these changes :) For the `qbs_to_string()` thing don't feel like you need to do that for this review, it was just a thought for the future (and might be a potential way to remove the `libqb.h` dependency).

internal/c/parts/gui/gui.cpp
a740g: That's a kickass idea. We'll need to put these conversion functions somewhere then. I'm thinking `qbs.h` & `qbs.cpp` in `internal/c/libqb/`. We'll also have to tie this to `build.mk`. Let me know what you think.

internal/c/parts/gui/gui.cpp
a740g: Yeah. This isn't going to be used anymore as InForm will not work with 3.7 anyway, because of the font changes.
I've already swapped out the calls for this with _MESSAGEBOX in the PE version of InFrom I'm working on.

internal/c/parts/gui/gui.cpp
a740g: Re-read your main comment. XD Yeah. It's best to leave the conversion stuff for a later as it requires some planning. Also, now that I'm thinking about, we'll need to do and use the conversions carefully to avoid copying strings as much as possible to avoid performance hits.

RhoSigma-QB64 (approved):

PR #346, opened 2023-06-09
Improves code export

RhoSigma-QB64:
- added ability to export as new [qb=export] Forum codebox
- Forum/Wiki exports now go to the clipboard and can directly be pasted into the Forum post or Wiki page

### DIFF
--- a/source/ide/ide_methods.bas
+++ b/source/ide/ide_methods.bas
@@ -20827,24 +20827,28 @@ SUB ExportCodeAs (docFormat$)
     IF cu% GOTO CustomNoLink
     GOSUB FindWikiPage
     IF me% AND le% AND co% THEN post% = 0: what$ = "co": GOSUB CloseText: co% = 0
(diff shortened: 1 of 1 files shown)


a740g (approved):

source/ide/ide_methods.bas
```diff
@@ -20604,29 +20620,38 @@ SUB ExportCodeAs (docFormat$)
...
-    ideerror = 7
-    OPEN idepath$ + idepathsep$ + pNam$ + ext$ FOR OUTPUT AS #151
-    ideerror = 1
-    PRINT #151, LEFT$(eTxt$, ePos& - 1);
-    CLOSE #151
     PCOPY 2, 3
+    SELECT CASE LCASE$(docFormat$)
+        CASE "html", "rich"
+            ideerror = 7
+            OPEN idepath$ + idepathsep$ + pNam$ + ext$ FOR OUTPUT AS #151
```
mkilgore, line 20627: Is there a reason not to use `FREEFILE` here? If we did then we wouldn't have to worry about 151 already being in use.

mkilgore (approved):

source/ide/ide_methods.bas
RhoSigma-QB64: Yep, found that file 151 is used for quick "Write in place" operations in the qb64pe sources, hence immediate OPEN-WRITE-CLOSE constructs like saving changed source, I just adapted to it, as writing the export file is practically the same operation.

PR #356, opened 2023-07-11
Savefile GUI added to IDE

SteveMcNeill:
Added the ability to use the Save Dialogs in the IDE.


SteveMcNeill:
Not tested on Linux/Mac, but I don't think this'll have any problems.  To test, just type in a simple Hello World program and then click the Save As option with the UseGUIdialogs option (whatever it's actually called) enabled.  It should propose a name, which you can ignore if you wish, and then save your file for you wherever you tell it to.

a740g (approved):

RhoSigma-QB64 (approved):

Issue #349, opened 2023-06-14
Graphical Error


TheXadGaming:
For me anyway, when I Fullscreen the IDE there's a black bar on the top of the screen, offsetting everything down:

![image](https://github.com/QB64-Phoenix-Edition/QB64pe/assets/108760829/d704315b-f735-4e37-a4b3-c462128b4dc5)

SteveMcNeill:
Some times, it's impossible for there to *NOT* be a black bar on the screen.   The IDE runs in SCREEN 0, so it's a text screen -- which means that it *has* to work with a monospaced font.  Every line has to be the same height and every character has to be the same width.

Now, understanding that basic rule and limitation of our interface, let's go with the following configuration for our PC:
Our desktop is 1000 pixels tall.  (This makes for easy maths and illustration of the point.)
We choose a font that is 10 pixels in height.   
Our Title Bar is set by the system to be 20 pixels in height.

Now, how does this all play together to create a screen for us?   We end up with a screen with a 20 pixel title bar and 98 rows of 10 pixel characters, with ZERO black space left over.  Everything worked out perfectly, life is good, and our IDE is using the full screen.

Now, let's change that to where the system sets the Title Bar to being 24 pixels high, but everything else is the same -- 1000 pixel desktop height, 10 pixel font height.  How's that work out?

We now have an IDE with a 24 pixel Title Bar and 97 rows of 10 pixels characters, with 6 pixels which we have to make a choice on:  Do we create a partial row with 4 pixels missing?  Or do we just leave those 6 pixels blank??  The Title Bar used 24 pixels, so we only have a total of 976 for the screen, and our font is 10 pixels in height.  Just what the heck do we do with the leftover pixels??

Now, someone will say, "But fullscreen doesn't have a Title Bar!"

Doesn't matter -- it still has the same inherent math involved.  Let's say our desktop is 1000 pixels in height.  Our font in 11 pixels in height.  11 * 90 = 990, with 10 pixels left over that we have to decide:  Partial row or blank row??

Nobody likes a partial row to text and buttons and such to interact with, so we chose a blank row for the IDE.  The math just doesn't allow to *NOT* have it, if on wants to maintain a true monospace font ratio with their display.

mkilgore:
@SteveMcNeill The problem is that the black bar is on the wrong side, it should be on the bottom but it's on the top.

[...]

Issue #313, opened 2023-03-15
Add Native Windows on ARM (ARM64) Support for QB64PE


a740g:
With the increasing popularity of Windows devices running on ARM processors, it would be great to add native Windows on ARM support for QB64PE. [Windows 11 on ARM](https://learn.microsoft.com/en-us/windows/arm/overview) can execute binaries for ARM, ARM64, x86 and x86-64 (the latter two use emulation).

Currently, QB64PE works on ARM-based devices running on Windows through emulation, which can be slow and inefficient. By adding native Windows on ARM support, QB64PE can take full advantage of the performance benefits of ARM processors, providing a better user experience for QB64PE users on these devices.

GCC MINGW does not currently support Windows on ARM. However, we have the [LLVM MinGW](https://github.com/mstorsjo/llvm-mingw) project that has support for targeting ARM / ARM64 on Windows. Although, this toolchain is based on LLVM / Clang, I was able to build the QB64PE IDE with minimal modifications.

We should also evaluate if [LLVM MinGW](https://github.com/mstorsjo/llvm-mingw) can be used as our backend for all platforms.

PR #358, opened 2023-07-29
Fix incorrect client RECT offset when compiled with LLVM-MingW

a740g:
When QB64-PE is compiled with LLVM-MingW, it offsets the Windows client RECT in such a way that part of it is hidden underneath the title bar. See image below. Also see https://qb64phoenix.com/forum/showthread.php?tid=1714

I tracked this down to the `fghComputeWindowRectFromClientArea_UseStyle()` function. Then, after comparing this to the [latest version](https://github.com/FreeGLUTProject/freeglut/blob/e9ceb6dd8df0b4b289b547e141ddb70b7f36cfe2/src/mswin/fg_window_mswin.c#L413) of FreeGLUT, I saw that the new implementation was much simpler. So, I replaced the old implmentation with the new one.

We should seriously consider moving to the latest version of FreeGLUT. I suspect that #349 is also related to the version of FreeGLUT that we are using.

This PR help us to move a little closer to achieving #313.

image

### DIFF
--- a/internal/c/parts/core/src/freeglut_window.c
+++ b/internal/c/parts/core/src/freeglut_window.c
@@ -989,31 +989,24 @@ void fgSetWindow ( SFG_Window *window )
  */
 void fghComputeWindowRectFromClientArea_UseStyle( const DWORD windowStyle, RECT *clientRect, BOOL posIsOutside )
 {
-    int xBorderWidth = 0, yBorderWidth = 0;
+    RECT windowRect = {0, 0, 0, 0};
+    DWORD windowExStyle = 0;
+    
+    CopyRect(&windowRect, clientRect);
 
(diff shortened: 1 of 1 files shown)


mkilgore (approved):

RhoSigma-QB64 (approved):

SteveMcNeill:
What's required to upgrade to the latest version of FreeGlut?  Is it just a simple swap out of the old version for the new one, or is there a lot of tweaking which would be required under the hood?

a740g:
Unfortunately, it's not going to be a simple swap. :(

@mkilgore and I talked about this on Discord. There are multiple custom modifications that have been made to the library in the past. We'll need to track all these changes and then check how to make it work with the swap without making invasive changes to the new library version.

PR #363, opened 2023-08-26
Adding clean.sh to complement clean.bat for Linux/Mac users

grymmjack:
`internal/clean.bat` is a nice thing to have in a pinch on Windows.

Adding the Linux/MacOS equivalent for those OSes.


mkilgore (commented):
This is the wrong approach. With the new `Makefile`-based build system that QB64-PE has this doesn't actually clear out all the built flles, but also `make clean` can already do all this for you, so we don't need a script for it.

Really what we should do is remove `clean.bat`, I thought I had already either gotten rid of all of those scripts or replaced their contents with a call to `make`, Ex: [purge_all_precompiled_content.sh](https://github.com/QB64-Phoenix-Edition/QB64pe/blob/main/internal/c/purge_all_precompiled_content.sh)

The one thing `make clean` won't do is the `source/` folder "pruning", but that really doesn't matter anyway, nobody should be touching `source/` by hand since it's updated via the automated process in the repo.

a740g:
@mkilgore @grymmjack I had noticed `internal/clean.bat` too before but never really bothered to check what was inside. The only reference to this file is in: https://github.com/QB64-Phoenix-Edition/QB64pe/blob/64f2384619a84345c04d163916676af1bed5fc27/.ci/make-dist.sh#L46

It just copies the file and does nothing with it.

We could get rid of this file and line in `.ci/make-dist.sh` using this PR? 🤷‍♂️

I always use `.\internal\c\c_compiler\bin\mingw32-make.exe OS=win clean` on Windows when I want to clean everything.

SteveMcNeill:
We've had this PR here for several months now.  Is this something left to pursue?  Something to drop?  We should clear our docket with the latest release and decide what's going on here -- if anything.  If nothing, then this needs to just be closed with a "Not at this time resolution".  ;)

a740g:
Closing this after checking with @grymmjack

PR #367, opened 2023-09-14
Support for PCF & BDF fonts

a740g:
This PR:
- Adds complete support for PCF and BDF bitmap fonts
- Fixes a macOS segfault reported by @grymmjack [here](https://qb64phoenix.com/forum/showthread.php?tid=1990&pid=19690#pid19690)


mkilgore (approved):
What was causing the seg-fault?

a740g:
> What was causing the seg-fault?

Me using `utf32.codepoints.data()` before calling `utf32.ConvertASCII()`. 🤦‍♂️ Stupid mistake. But I'm not sure why it never failed on Windows / Linux. Probably because of the underlying `std::vector` implementation. 🤷‍♂️

mkilgore:
> Me using `utf32.codepoints.data()` before calling `utf32.ConvertASCII()`. 🤦‍♂️ Stupid mistake. But I'm not sure why it never failed on Windows / Linux. Probably because of the underlying `std::vector` implementation. 🤷‍♂️

Ouch, I did not catch that at all, I thought that was just a readability change. IIRC the evaluation order of function arguments isn't specified and the compiler gets to pick so maybe that's why it works on Linux/Windows but not on MacOS, we just got lucky with the order.

RhoSigma-QB64 (approved):

grymmjack:
Sweet! Glad this is fixed. When it's ready for testing I can try it on the stuff that was segfaulting to confirm on my mac @a740g unless you've already done this, in which case thank you thank you (2 thanks vs. just the one. :P)

a740g:
> Sweet! Glad this is fixed. When it's ready for testing I can try it on the stuff that was segfaulting to confirm on my mac @a740g unless you've already done this, in which case thank you thank you (2 thanks vs. just the one. :P)

Yeah. I tested this on macOS. Actually, I did the whole debugging and editing on macOS. It's a little frustrating because of the whole keyboard shortcut situation but I think I am getting used to it. 😊

PR #374, opened 2023-09-22
Remove OpenAL audio backend

a740g:
This PR removes the legacy OpenAL based audio backend from QB64-PE.

See https://github.com/QB64-Phoenix-Edition/QB64pe/milestones


RhoSigma-QB64 (approved):

source/qb64pe.bas
```diff
@@ -13334,9 +13319,6 @@ FUNCTION ParseCMDLineArgs$ ()
                 token$ = MID$(token$, 3)
 
                 SELECT CASE LCASE$(LEFT$(token$, INSTR(token$, "=") - 1))
-                    CASE ":useminiaudio"
```
mkilgore, line 13337: This is an interesting case, I'd be tempted to still allow `useminiaudio=true` from a backwards compatibility perspective. But realistically nobody is using this so let's just take it out :D

source/subs_functions/subs_functions.bas
```diff
@@ -233,11 +233,12 @@ id.hr_syntax = "_MEMIMAGE or _MEMIMAGE(imageHandle)"
 regid
 
 clearid
-id.n = qb64prefix$ + "MemSound": id.Dependency = DEPENDENCY_AUDIO_DECODE
+id.n = qb64prefix$ + "MemSound": id.Dependency = DEPENDENCY_MINIAUDIO
 id.subfunc = 1
 id.callname = "func__memsound"
 id.args = 2
 id.arg = MKL$(LONGTYPE - ISPOINTER) + MKL$(LONGTYPE - ISPOINTER)
+id.specialformat = "?[,?]"
```
mkilgore, line 241: Huh, was this a bug that the second parameter wasn't optional?

mkilgore (approved):

source/qb64pe.bas
a740g: BTW, thanks. Your documentation (QB64pe/docs) made this whole thing a breeze for me. I picked up a couple of useful things from there. Top stuff!

source/subs_functions/subs_functions.bas
a740g: Yeah. It wasn't. We should have made it optional the first time. We made it optional with this PR. See the discussion here: https://discord.com/channels/975381912350752819/975389657212805160/1154682098913910844

PR #380, opened 2023-10-01
Initial WoA64 and MingW-LLVM support

a740g:
This PR adds initial support for Windows on ARM (WoA) using the MingW-LLVM toolchain. While this change does not currently impact the CI process, it enables users to build a native binary of QB64-PE on WoA64 devices by utilizing the `setup_win.cmd` script (which fetches MingW-LLVM when a WoA environment is detected).

- This update lays the foundation for compiling QB64-PE on Windows on ARM (WoA64) devices
- The MingW-LLVM toolchain is used to facilitate the compilation process specifically for WoA64
- Users can now clone the repository on their WoA64 device and execute `setup_win.cmd` to automatically retrieve MingW-LLVM and build QB64-PE as a native binary
- It's important to note that this change does not impact our CI workflow

See #313


internal/c/libqb.cpp
```diff
@@ -32199,10 +32199,18 @@ int32 addone(int32 x) { return x + 1; } // for testing purposes only
 qbs *func__os() {
     qbs *tqbs;
 #ifdef QB64_WINDOWS
-#    ifdef QB64_32
+#   ifdef QB64_32
+#       ifdef QB64_ARM
+    tqbs = qbs_new_txt("[WINDOWS][32BIT][ARM]");
```
mkilgore, line 32204: You can have the compiler concatenate string constants for you by just placing them one-after-another, so I think we can replace all this logic with something like this:

```c
#ifdef QB64_WINDOWS
#    define QB64_OS_SYSTEM_STR "[WINDOWS]"
#elif defined(QB64_LINUX)
#    define QB64_OS_SYSTEM_STR "[LINUX]"
#elif defined(QB64_MACOSX)
#    define QB64_OS_SYSTEM_STR "[MACOSX][LINUX]"
#else
#    define QB64_OS_SYSTEM_STR ""
#endif

#ifdef QB64_32
#    define QB64_OS_BITS_STR "[32BIT]"
#else
#    define QB64_OS_BITS_STR "[64BIT]"
#endif

#ifdef QB64_ARM
#    define QB64_OS_ARCH_STR "[ARM]"
#else
#    define QB64_OS_ARCH_STR ""
#endif

    // Have the compiler combine all our selections into one string
    tqbs = qbs_new_txt(QB64_OS_SYSTEM_STR QB64_OS_BITS_STR QB64_OS_ARCH_STR);
```

mkilgore (approved):

internal/c/libqb.cpp
a740g: Done.

RhoSigma-QB64 (approved):

PR #392, opened 2023-10-09
$VERSIONINFO enhancement (#20)

### LINKED ISSUE #20: $VERSIONINFO should accept a quoted string
It's not clear to me why exactly this is the case, but currently the various $VERSIONINFO values accept non-quoted string information. Ex:

`$VERSIONINFO:ProductName=foo bar`

I personally find this extremely strange, as far as I know this is the only command that accepts such a thing. To me it seems like it should accept a quoted string (single quotes I think, since that's what other metacommands do), Ex:

`$VERSIONINFO:ProductName='foo bar'`

This could be done in a largely backwards compatible way as long as we fallback to the current behavior if the line doesn't begin and end with a quote (though it would be annoying that such behavior doesn't just produces an error, rather than expected behavior). To avoid that issue perhaps the current syntax could produce a warning, but I don't know much about the warning system yet. Also, there is a slight backwards compatibility break as that line currently would place the literal "'foo bar'" into the version information, but I don't think that's a big enough deal to worry about - I think most who wrote it that way probably didn't expect the quotes to be in the output to begin with.

a740g:
This PR implements `$VERSIONINFO` enhancement as described in #20.

This works in a backwards compatible way which allows the old behavior but also suggests the user (via warnings) to enclose the strings using the string bracket delimiter `''`.

- Allows old behavior
- `FILEVERSION#` and `PRODUCTVERSION#` are exempted
- Honors warning setting - `Options > Ignore Warnings`

Warning messages for current `qb64pe.bas` `$VERSIONINFO`:
image

Closes #20.


RhoSigma-QB64:
Shouldn't this PR not do the required addjustments in qb64pe.bas to avoid those warnings?

RhoSigma-QB64:
For me it's unclear why we need to throw warnings at all, if both methods (w/ and w/o quotes) are valid and allowed? But, if we wanr about missing quotes, then we should clean our own house, making sure we don't get warnings on our own code. Just a matter of my sick perfectionism I guess :)

a740g:
> For me it's unclear why we need to throw warnings at all, if both methods (w/ and w/o quotes) are valid and allowed? But, if we want warn about missing quotes, then we should clean our own house, making sure we don't get warnings on our own code. Just a matter of my sick perfectionism I guess :)

Ah. Ok. I get it. And you are right.

I added the warnings because that's what #20 suggested. And TBH, adding warnings does seem a little useful. Without the quotes the IDE attempts to syntax-highlight the string which ends up looking ugly. The warning basically motivates the user to add the quotes.

Fun fact: When I first started using QB64, I always used to add the quotes to the `$VERSIONINFO` strings thinking that was a natural thing to do.

I am good both ways:
1. We can drop the warnings completely, or
2. We just add the quotes to clean our own house

RhoSigma-QB64:
> Without the quotes the IDE attempts to syntax-highlight the string which ends up looking ugly. The warning basically motivates the user to add the quotes.

Yeah, noticed that too when implementing the code export stuff, but it's rather a matter of finetuning the syntax highlighter then. I already did that for the code export to avoid this faulty highlighting. I can look at the IDE highlighter later, for now I'd say just clean our house :)

a740g:
> > Without the quotes the IDE attempts to syntax-highlight the string which ends up looking ugly. The warning basically motivates the user to add the quotes.
> 
> Yeah, noticed that too when implementing the code export stuff, but it's rather a matter of finetuning the syntax highlighter then. I already did that for the code export to avoid this faulty highlighting. I can look at the IDE highlighter later, for now I'd say just clean our house :)

Done.

RhoSigma-QB64 (approved):

RhoSigma-QB64:
BTW in #20 Matt mentioned that by his knowlegde $VERSIONINFO was the only Meta-Command which allows unquoted strings, but I know at lest one other, it's $ERROR, maybe you want too look at that too and fix it in the same run.

a740g:
> BTW in #20 Matt mentioned that by his knowlegde $VERSIONINFO was the only Meta-Command which allows unquoted strings, but I know at least one other, it's $ERROR, maybe you want to look at that too and fix it in the same run.
> 
> What's about $LET, can we define a string such as $LET me = RhoSigma? Must admit I avoid the pre-compiler till now.

Fair callout on `$ERROR`. I'll get that to be compliant as well.
I usually use `$LET` to create per-processor variables and seldom assign any values. It however requires strings to be properly quoted using `"` if you are assigning one.
While on the topic, `$MIDISOUNDFONT` seems to require `"` instead of `'`. I am not sure if this is intentional and requires any change. 🤷‍♂️

a740g:
Enabled $ERROR to handle strings enclosed using `''` as suggested by @RhoSigma-QB64.

RhoSigma-QB64:
With your new FUNCTION RemoveStringEnclosingPair() you may want to check its usability for `'$INCLUDE: 'file.bm'` too have not several different handlings for several meta-commands.

a740g:
> With your new FUNCTION RemoveStringEnclosingPair() you may want to check its usability for `'$INCLUDE: 'file.bm'` too have not several different handlings for several meta-commands.

I looked at the `$INCLUDE` handling [here](https://github.com/QB64-Phoenix-Edition/QB64pe/blob/1d95c731cfb1810c6f404cca92705cc9521da0ed/source/qb64pe.bas#L20677) and it seems like there is special handling been done due to this metacommand being embedded in a comment and some QB45 behavior. It does mandate usage of `'` to enclose file names. This is part of the giant `lineformat$()` function. IMO, it is best we leave it the way it is and re-visit it sometime in the future as part of a significant compiler re-write / simplification.

Issue #224, opened 2022-10-28
CONST not evaluating properly when using a literal


ghost:
When using the below code with the latest source (clone and run setup_win.bat), this CONST expression does not work. It did work in 1.5. If I replace the `&HFFF` literal with a new constant that is set to that value then the IDE stops complaining.
```
    Const STANDARD_RIGHTS_REQUIRED = &H000F0000
    Const SYNCHRONIZE = &H00100000
    Const PROCESS_ALL_ACCESS = STANDARD_RIGHTS_REQUIRED Or SYNCHRONIZE Or &HFFF
```

mkilgore:
It turns out it's quite simple, but also probably a mess to fix. Basically in `PreParse()` in the "PP_TypeMod" step all type suffixes are replaced with a "C_*" equivalents which is used to (I guess) make handling them easier later by turning them into a kind of function. The big issue is that the logic can't tell that the `&` is not a type suffix, so it gets replaced with `C_LO` which obviously then breaks everything after.

The single `&H00100000` cases work because while it bombs out of the `Evaluate_Expression$` logic, the particular way they bombs out allows it to get handled by the `fixoperationorder$` and `evaluateconst$` logic which can handle that expression and produces the correct value. When the expression is after an operator like `Or` then the resulting string bombs out of the `Evaluate_Expression$` logic with an error that stops the `CONST` evaluations and never calls `evaluateconst$`.

Sadly my const test suite is missing cases for `&H` and friends, or else this probably would have been noticed. Cases can be easily added at least.

mkilgore:
Fixed with #397

Issue #193, opened 2022-09-25
CONST replaces the wrong CONST variable in some circumstances


mkilgore:
The following code is valid but evaluates wrong, it should produce `3` but it actually produces `5`:

```vb
CONST foo = 2
CONST bar2 = 3
CONST fooORbar2 = 5
CONST foobar = foo OR bar2

PRINT foobar
```

The problem is that `PreParse` in the `Evaluate_Expression$` logic strips spaces (same problem as #192), which can cause some valid expressions to be transformed into a different valid expression. In this case, removing spaces transforms the expression into `fooORbar2`, which is the same name as an existing CONST variable. The CONST variables are then replaced by simple string replacement, so the `fooORbar2` CONST gets found and 5 is replaced, rather than properly using `foo` and `bar2` for the replacement. More complex expression could presumably produce stranger replacements, since the only real requirement is that the name of an existing const exists in the expression after spaces are removed.

mkilgore:
Fixed with #397

PR #397, opened 2023-10-15
Patch to CONST

SteveMcNeill:
Should fix both of these issues:
    https://github.com/QB64-Phoenix-Edition/QB64pe/issues/193 
    https://github.com/QB64-Phoenix-Edition/QB64pe/issues/224

### DIFF
--- a/source/qb64pe.bas
+++ b/source/qb64pe.bas
@@ -24708,6 +24708,67 @@ SUB PreParse (e$)
     DIM f AS _FLOAT
     STATIC TotalPrefixedPP_TypeMod AS LONG, TotalPP_TypeMod AS LONG
 
+    DO 'convert &H values to decimal values to prevent errors
+        l = INSTR(l, UCASE$(e$), "&H")
+        IF l THEN
+            FOR l1 = l + 2 TO LEN(e$)
+                SELECT CASE UCASE$(MID$(e$, l1, 1))
+                    CASE "0" TO "9"
+                    CASE "A" TO "F"
+                    CASE ELSE:
+                        PRINT UCASE$(MID$(e$, l1, 1))
+                        EXIT FOR
+                END SELECT
+            NEXT
+            IF l1 <> l + 2 THEN 'hex number found
+                IF l1 > l + 18 THEN EXIT DO
+                l$ = LEFT$(e$, l - 1)
+                r$ = MID$(e$, l1)
+                t~&& = VAL(MID$(e$, l, l1 - l))
+                m$ = _TRIM$(STR$(t~&&))
+                e$ = l$ + m$ + r$
+            ELSE
+                EXIT DO
+            END IF
+        END IF
+    LOOP UNTIL l = 0
+
+    'replace existing CONST values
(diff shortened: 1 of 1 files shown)


source/qb64pe.bas
```diff
@@ -24708,6 +24708,67 @@ SUB PreParse (e$)
...
 
+    DO 'convert &H values to decimal values to prevent errors
+        l = INSTR(l, UCASE$(e$), "&H")
+        IF l THEN
+            FOR l1 = l + 2 TO LEN(e$)
+                SELECT CASE UCASE$(MID$(e$, l1, 1))
+                    CASE "0" TO "9"
+                    CASE "A" TO "F"
+                    CASE ELSE:
+                        PRINT UCASE$(MID$(e$, l1, 1))
```
mkilgore, line 24719: We probably don't want to `PRINT` :D

source/qb64pe.bas
```diff
@@ -24708,6 +24708,67 @@ SUB PreParse (e$)
...
+                        EXIT FOR
+                END SELECT
+            NEXT
+            IF l1 <> l + 2 THEN 'hex number found
+                IF l1 > l + 18 THEN EXIT DO
```
mkilgore, line 24724: Should this return back an error instead?

(Same question about the other `EXIT DO` as well)

source/qb64pe.bas
mkilgore, line 1: I can't add a comment on it, but down on line 24927 is some old logic for the `&H` replacement, I'm assuming it could be removed.

Additionally I think it'd be good to expand the [`expression.bas`](https://github.com/QB64-Phoenix-Edition/QB64pe/blob/main/tests/compile_tests/const/expression.bas) tests for `CONST` to cover some of these cases involving variables and `&H`, I didn't add any of those. If you feel like it you could also add some error cases for stuff like `&H`, `&HXX`, etc.

mkilgore (approved):

source/qb64pe.bas
SteveMcNeill: Debugging statement missed.  LOL.  Good catch.  :)

source/qb64pe.bas
SteveMcNeill: It does.  Give it a try:   ``PRINT &H12345678901234567``

Overflow Error on Line ###.

source/qb64pe.bas
SteveMcNeill: If I'd only noticed that code before writing the patch!  It does basically the same thing which I'm now doing -- it's just doing it after the suffix replacement, rather than before it.  I imagine all that was needed for this to work as intended, is to just move this to the start of the SUB, rather than it being one of the last things we process. 📦 

As it is, it's redundant and good to strip out now.

source/qb64pe.bas
SteveMcNeill: Fixed

source/qb64pe.bas
SteveMcNeill: Cleaned up and removed.

PR #398, opened 2023-10-15
Fix to Mod and \ in CONST

SteveMcNeill:
As per the sample code below:

Const a = 5 Mod 2.8
Const b = 5 \ 2.8
Print a; "="; 5 Mod 2.8
Print b; "="; 5 \ 2.8

Old output was wrong.  Patch should make the CONST values the same as what QB64 normally generates for us.


a740g (approved):

mkilgore (approved):
LGTM :)

A bit of history: This bug was introduced by me in https://github.com/QB64-Phoenix-Edition/QB64pe/pull/126, a fix for https://github.com/QB64Official/qb64/issues/4. The issue here was then reported in https://github.com/QB64Official/qb64/issues/53 (and noticed by us) and recently fixed over in QB64Official in https://github.com/QB64Official/qb64/pull/54 by simply reverting my old change and reintroducing the bug in https://github.com/QB64Official/qb64/issues/4.

Using `_Round()` instead of `FIX()` is the proper solution as it matches what QB64 does when converting the `Double` argument to `Mod` into an integer, ensuring that the `= 0` comparison is accurate while not changing the behavior of `Mod`.

PR #401, opened 2023-10-18
Add Support for optional paramter for an image handle to CLS

SteveMcNeill:
Without having to redo a whole crap ton of old code to take the new parameter, this seems like the easiest way to implement the addition of a new parameter to CLS for an optional image handle.

Tested with the simple code below, which runs and works as expected, with no issues:
```
$Color:32
Screen _NewImage(640, 480, 32)
foo = _NewImage(480, 320, 32)

_Dest foo
Cls , Red
_Dest 0
Cls , Blue

'at this point, both screens should be colored and our Dest is the 640x480 screen
Cls , Green, foo
Sleep
_PutImage (0, 0), foo 'If the CLS worked, we should put a green screen on the blue one, instead of a red one.
```


source/subs_functions/subs_functions.bas
```diff
@@ -2833,11 +2833,11 @@ regid
...
 id.subfunc = 2
-id.callname = "sub_cls"
-id.args = 2
-id.arg = MKL$(LONGTYPE - ISPOINTER) + MKL$(ULONGTYPE - ISPOINTER)
-id.specialformat = "[?][,?]"
-id.hr_syntax = "CLS [method%] [, bgColor&]"
+id.callname = "sub_clsDest"
+id.args = 3
+id.arg = MKL$(LONGTYPE - ISPOINTER) + MKL$(LONGTYPE - ISPOINTER) + MKL$(ULONGTYPE - ISPOINTER)
+id.specialformat = "[?][,?][,?]"
```
mkilgore, line 2839: Shouldn't it be something like `[?][,[?][,?]]`?

internal/c/libqb.cpp
```diff
@@ -14684,6 +14685,14 @@ void qbg_sub_view(int32 x1, int32 y1, int32 x2, int32 y2, int32 fillcolor, int32
 }
 
 void qbg_sub_locate(int32 row, int32 column, int32 cursor, int32 start, int32 stop, int32 passed);
+
+void sub_clsDest(int32 method, uint32 use_color, int32 dest, int32 passed) {
+    int32 tempDest;
+    tempDest = func__dest();
+    sub_cls(method, use_color, passed);
```
mkilgore, line 14692: Isn't a call to `sub__dest(dest)` missing here?

Also, how does the potential error from `sub__dest(dest)` for an invalid handle end up working?

mkilgore (commented):
As always, automated tests would be good :P Your example could pretty easily be turned into an automated image test like these [PAINT](https://github.com/QB64-Phoenix-Edition/QB64pe/blob/main/tests/compile_tests/paint/tile_border_black_bg.bas) ones I made.

source/subs_functions/subs_functions.bas
mkilgore: Sorry, my point was that the syntax `[?][,?][,?]` implies that you can include the third parameter without the comma for the second. The syntax `[?][,[?][,?]]` requires that to use the third argument you have to include the comma for the second.

Ex. [_LOADIMAGE](https://github.com/QB64-Phoenix-Edition/QB64pe/blob/e84c80e5ff511d205af661a2636782fd04a1f4b6/source/subs_functions/subs_functions.bas#L1066) with two optional parameters is written this way.

internal/c/libqb.cpp
a740g: Ok. I am late to the party. 😊

But why two `cls` functions? Can't we have the whole thing implemented in `sub_cls()`?

mkilgore: I think he did that due to the build issues with removing `sub_cls()` outright or changing it's parameters. I'd agree I'd rather just have one though, and then maybe an internal `cls()` function for the `libqb.cpp` callers.

Doing it that way in a single step would probably require editing `./internal/source` to make it call the new `sub_cls()`, so not tons of fun.

SteveMcNeill: because that breaks the world, as we were talking about in Discord the other day.  all the libqb functions want to make use of the already existing version with 3 params -- they'd all need to be changed.  QB64 won't compile itself, as it was built with the 3 param version of CLS, not 4, so you get argument mismatch issues...  

Honestly, the easiest way just seems to have a 2nd command, which QB64 can call, without affecting all the old calls that libqb (and who knows what all other files) are making to the old version without that extra param.

a740g: Ah ok. Got it. We should be able to flatten it with a future update.

SteveMcNeill:
```
internal\c/../temp/main.txt:144986:8: error: too few arguments to function 'void sub_cls(int32_t, uint32_t, int32_t, int32_t)'
144986 | sub_cls(NULL,NULL,0);
```

^ Those style glitches.

source/subs_functions/subs_functions.bas
```diff
@@ -2833,11 +2833,11 @@ regid
...
 id.n = "Cls"
 id.subfunc = 2
-id.callname = "sub_cls"
-id.args = 2
-id.arg = MKL$(LONGTYPE - ISPOINTER) + MKL$(ULONGTYPE - ISPOINTER)
-id.specialformat = "[?][,?]"
-id.hr_syntax = "CLS [method%] [, bgColor&]"
+id.callname = "sub_clsDest"
+id.args = 3
+id.arg = MKL$(LONGTYPE - ISPOINTER) + MKL$(LONGTYPE - ISPOINTER) + MKL$(ULONGTYPE - ISPOINTER)
```
mkilgore, line 2838: It probably doesn't significantly matter, but I think it should probably be `LONG/ULONG/LONG` to keep the color type as an unsigned long. Your current change makes it `LONG/LONG/ULONG`.

SteveMcNeill: Swapped those around.  In libqb, I had the order correct; in the bas source, they were reversed.  Only 27 and a half places necessary to track such minor changes...  LOL, and I had to retweak 26 of those!  :D

RhoSigma-QB64 (approved):

Issue #362, opened 2023-08-26
Max number of subs and functions in entire program limited to `1000`


grymmjack:
It would be nice to have the ability to exceed `1000`

I created a simple test here, changing `1000` to `10000` in `qb64pe.bas`:
```vb
933: DIM SHARED sfidlist(10000) AS LONG
934: DIM SHARED sfarglist(10000) AS INTEGER
935: DIM SHARED sfelelist(10000) AS INTEGER
```

And it works.

My code which exceeded the `1000` limit was discovered because it was showing a compiler error with line number `11756`.

The code from `qb64pe.bas` around lines `11756` is this:
https://github.com/QB64-Phoenix-Edition/QB64pe/blob/4c50c8db3cb33c52b218d508583c7ff82571a838/source/qb64pe.bas#L11729-L11765

Thanks @mkilgore for your help in discord. I think just 10x'ing the limit would fix the problem but you mentioned you want to be more frugal as you said:
```
Well typically for these we just redim them to be larger after checking the size, we could do that pretty easily here I think.
```

mkilgore:
I think the redim approach is better long-term, but if it's a bit too messy to implement then raising the limit to 10000 seems fine to me, the increase in memory usage is negligible 🤷 

That said, if we're not doing the `redim` approach then I think I'd like to see a simple error check so that the user gets a better error about reaching the max number of subs/functions (instead of an internal compiler error). But at that point, you could just replace the error message with a `redim` and fix the problem, so perhaps it's not worth it XD...

SteveMcNeill:
I think this needs a couple of things to fix it properly.
Redim and increase the limit by 1000 per increase, but only to a finite hard limit.  The reason for this limit is in case the IDE glitches at some point translating the BAS to C, to prevent an endless loop using up all one's memory and swap file space.
When the hard limit is hit, the IDE needs to toss a more descriptive error message asking the user to report the glitch and share code via discord, github, or the forums with the dev team.  In *MOST* cases, an user isn't going to have 10,000 subs and functions in a program.  (Currently 1000, but the actual limit will be whatever the Redim cap tops out at.)  For most users,if there's a glitch here, it's on our end and not there's, and is something we need to be aware of to address.

grymmjack:
Thanks @a740g for this fix!

SteveMcNeill:
Yeah...  *sniffle*   Thanks for this fix @a740g....   

[...]

PR #404, opened 2023-10-30
Increase Sub/Function Id Limit

SteveMcNeill:
As per #362, this raises the limit in increments from 1,000 to a maximum of 25,000 subs and functions for a single QB64PE program.  If that number is exceeded, the IDE will give an error message. reporting the issue for the user.


source/qb64pe.bas
```diff
@@ -11767,6 +11767,16 @@ FOR i = 1 TO idn
 
                             unresolved = unresolved + 1
                             sflistn = sflistn + 1
+                            ubound_sf = UBOUND(sfidlist) 'all 3 should have the same limit
+                            IF sflistn > ubound_sf THEN
+                                REDIM _PRESERVE sfidlist(ubound_sf + 1000) AS LONG
+                                REDIM _PRESERVE sfarglist(ubound_sf + 1000) AS INTEGER
+                                REDIM _PRESERVE sfelelist(ubound_sf + 1000) AS INTEGER
+                            END IF
+                            IF ubound_sf > 25000 THEN 'manually set a descriptive error message for the user so they know what's happening.
```
RhoSigma-QB64, line 11776: Should this rather be `IF sflistn > 25000 THEN` instead of `ubound_sf` ?? Also better move the error checking block before the REDIM block.

As it currently is, it would unnecessarily REDIM the arrays to 26000 elements.

SteveMcNeill: Good catch.   And with the misuse of ubound_sf there, it'd never have a chance to STOP erroring out.   sflistn resets itself and rebuilds itself with each new pass of the IDE (so if you insert a SUB between subs, it'll take the old id numbers from them and try and keep sub numbering in order), so when someone removes a SUB, that value *should* drop by one, in total.   The idea here is for the IDE to just toss an error message, without crashing, and which the user could then consolidate subs/functions (or remove unneeded ones), to get back below that 25,000 limit and stop the error from ever occurring.

Honestly though, I can't hardly imagine that anyone would *ever* have 25,000 subs and functions in a program.  From the day QB64 was first conceived, to now, Grymm has been the only person to ever report the glitch with a cap of just 1,000.   I truly think if that cap is ever reached, it's going to be because of some endless loop on our part in the IDE, and not something the user can ever fix to begin with.  :)

RhoSigma-QB64 (approved):

a740g (approved):

PR #405, opened 2023-10-31
Patch _ARCSEC and _ARCCSC

SteveMcNeill:
As per the forum posts here: https://qb64phoenix.com/forum/showthread.php?tid=2125

I'm not the math guru that I used to be, but it simply seems as if the error checking for invalid domains on these were basically just reversed.  In cases where we should've tossed an error, we were giving an answer.  In cases where we should have given a valid answer, we were tossing an error.  

This should patch the issue for both the affected commands.


internal/c/libqb.cpp
```diff
@@ -33924,7 +33924,7 @@ double func_pi(double multiplier, int32 passed) {
 
 double func_arcsec(double num) {
     int sign = (num > 0) - (num < 0);
-    if (num < -1 || num > 1) {
+    if (std::abs(num) < 1) {
```
mkilgore, line 33927: Are we sure the formula isn't wrong too? The check appears to avoid negatives in the `std::sqrt(1 - num * num)` calculation, so in that respect the original constants were correct. Since we know the constraints are actually wrong, I would assume the formula must be wrong too and `std::sqrt(1 - num * num)` is not correct.

a740g: I think these are overly complicated.

I did some tests and derived both functions with a much simpler approach and tested the results as well.
See https://www.godbolt.org/z/c9Ys6P4j4

Basically, I used:

```c
// https://en.neurochispas.com/calculators/arcsec-calculator-inverse-secant-degrees-and-radians/
inline double func_arcsec(double num) {
    if (std::abs(num) < 1.0) {
        error(5);
        return 0.0;
    }

    return std::acos(1.0 / num);
}

// https://en.neurochispas.com/calculators/arccsc-calculator-inverse-cosecant-degrees-and-radians/
inline double func_arccsc(double num) {
    if (std::abs(num) < 1.0) {
        error(5);
        return 0.0;
    }
    
    return std::asin(1.0 / num);
}
```

Those `inline`s makes a world of difference and I think we should use them.

a740g: This is nice. I like it.

Use the `constexpr` keyword for `func_pi` so that compiler evaluates that to a constant at compile time. That will make QB64-PE math that uses it to run even faster by avoid the branch instructions.

```c
inline constexpr double func_pi(double multiplier, int32_t passed) {
    if (passed) {
        return 3.14159265358979323846264338327950288419716939937510582 * multiplier;
    }
    return (3.14159265358979323846264338327950288419716939937510582);
}
```

We can also include the `constexpr` keyword for the rest of the functions as which does not call std math functions. Like:
```text
func_deg2rad
func_rad2deg
func_deg2grad
func_grad2deg
func_rad2grad
func_grad2rad
```

Sorry for the churn. 😝

mkilgore (approved):

a740g (approved):

PR #407, opened 2023-11-07
Update QB64PE for faster file saving

SteveMcNeill:
As per the discussion here: https://qb64phoenix.com/forum/showthread.php?tid=2144


source/ide/ide_methods.bas
```diff
@@ -12458,12 +12458,19 @@ END FUNCTION
 
 SUB idesave (f$)
     ideerror = 6
-    OPEN f$ FOR OUTPUT AS #151
+    OPEN f$ FOR OUTPUT AS #151: close #151
+                              OPEN f$ FOR BINARY AS #151
     ideerror = 1
```
a740g, line 12463: We can probably set the line ending only once outside the loop so that we do not need to check repeatedly inside?

```vb
IF INSTR(_OS$, "WIN") then
    saveLineEnding = CHR$(13) + CHR$(10)
ELSE
    saveLineEnding = CHR$(10)
END IF
```

And then inside the loop we could just do:
```vb
outfile$ = outfile$ + a$ + saveLineEnding
```

a740g (approved):

RhoSigma-QB64 (approved):

PR #412, opened 2023-12-04
Avoid segfault on bogus SVG data

a740g:
This PR avoids a segfault that happens when binary data pretending to be valid SVG text is passed to the SVG parser.

For example:
```vb
img& = _LOADIMAGE("mysvg.svg", 32)  ' here if mysvg.svg has invalid binary junk it will trigger the segfault
```

For memory loads we check the data byte by byte while copying the data for `nsvgParse()`.
For file loads we check the file extension and check the data before passing it to `nsvgParse()`.


internal/c/parts/video/image/image.cpp
```diff
@@ -199,16 +242,24 @@ static uint32_t *image_svg_load_from_memory(const uint8_t *buffer, size_t size,
     if (!svgString)
         return nullptr;
 
-    memcpy(svgString, buffer, size);
-    svgString[size] = '\0';
+    // Bail if we have binary data. We'll also copy the data while doing the check to avoid another pass
+    for (size_t i = 0; i < size; i++) {
+        auto c = svgString[i] = buffer[i];
```
mkilgore, line 247: Note that `char` is typically signed so the +127 unicode values would trigger the `c < 32` check (I think the `auto` will be `char` due to the type of `svgString`). You should probably specify unsigned here if we want to allow them.

mkilgore (commented):
@a740g do we have any sense of where in nanosvg it blows up with a seg-fault?

I think ideally we'd address the issue in there, but only if it's simple enough. If it's a more fundamental problem with nanosvg then this type of approach makes sense, but we may want to consider looking for a replacement for nanosvg that correctly validates the input 🤷‍♂️

internal/c/parts/video/image/image.cpp
a740g: You are correct. Nice catch!

a740g:
> @a740g do we have any sense of where in nanosvg it blows up with a seg-fault?
> 
> I think ideally we'd address the issue in there, but only if it's simple enough. If it's a more fundamental problem with nanosvg then this type of approach makes sense, but we may want to consider looking for a replacement for nanosvg that correctly validates the input 🤷‍♂️

I am checking on this one now. It usually happens inside `nsvgParse()`.

SteveMcNeill (approved):

RhoSigma-QB64 (approved):

mkilgore (approved):

PR #416, opened 2023-12-16
Revert "Automatic update of ./internal/source"

RhoSigma-QB64:
This reverts commit ec12524d2b66b8dd10cbdeb81c86caeb4dc50412.


source/qb64pe.bas
```diff
@@ -3637,23 +3637,34 @@ DO
     'also check if such a handle was defined and mark it as used then.
     IF INSTR(a3u$, qb64prefix$ + "EMBEDDED$") > 0 THEN
         spo = INSTR(a3u$, qb64prefix$ + "EMBEDDED$"): ket = 0
+        ppo = INSTR(spo + 1, a3u$, "("): IF ppo = 0 THEN ppo = LEN(a3u$) + 1
```
mkilgore, line 3640: Testing locally, this doesn't catch the `Function get_embedded$(a$)` type of case.

source/qb64pe.bas
```diff
@@ -3637,23 +3637,34 @@ DO
...
+            'ignore everything in strings or after comment (' or REM)
             rsq = 0 'quote flag
             FOR i = ket + 1 TO spo
                 IF ASC(a3u$, i) = 34 THEN rsq = NOT rsq
                 IF ASC(a3u$, i) = 39 AND NOT rsq THEN EXIT DO
-                IF MID$(a3u$, i, 3) = "REM" AND NOT rsq THEN EXIT DO
+                IF MID$(a3u$, i, 3) = "REM" AND NOT rsq THEN
+                    IF i > 1 THEN rsf = ASC(a3u$, i - 1): ELSE rsf = 32 'front char
+                    rsr = ASC(a3u$, i + 3) 'rear char
+                    IF (rsf = 9 OR rsf = 32 OR rsf = 58) AND (rsr = 9 OR rsr = 32) THEN EXIT DO 'is real REM
```
mkilgore, line 3653: I don't know what the full rules are supposed to be, but it seems like as long as `REM` is in the right place then the character after REM can be almost anything. Ex: `rem$ _embedded$("foo")`, `rem; _embedded$("foo")`.

source/qb64pe.bas
```diff
@@ -3637,23 +3637,34 @@ DO
...
+                END IF
             NEXT i
+            IF rsq THEN EXIT DO 'inside a quoted string
```
mkilgore, line 3656: Along with the `skipElfCheck` jump before the loop, exiting the loop seems to cause the rest of the line to get ignored, even if there's more instances to check. Ex. In lines like `Print "_Embedded$"; _Embedded$("foo")` and `Print "_Embedded$("; _Embedded$("foo")` the second `_Embedded$` call is not getting checked even though it's real.

source/qb64pe.bas
```diff
@@ -3685,6 +3696,7 @@ DO
             'check for further occurrences
             spo = INSTR(spo + 2, a3u$, qb64prefix$ + "EMBEDDED$")
```
mkilgore, line 3697: I think you need to add the same check you do before loop again down here. Lines like `Print _Embedded$("foo"); var_embedded$` trigger the check on the `var_embedded$` when they shouldn't.

source/qb64pe.bas
RhoSigma-QB64: Right, it needs a check to the left side to and also must go inside the loop.

source/qb64pe.bas
RhoSigma-QB64: Obviously true, I just tested in the IDE when the REM gets highlighted and the line colored as comment, but that's not in line with when REM avoids compiling the code, so we have definitly an issue with the highlighter here too. Everything after REM is valid except A-z/0-9.

source/qb64pe.bas
RhoSigma-QB64: Sure, as above, the check must go inside the loop.

mkilgore (approved):

PR #419, opened 2023-12-28
Filesystem refactor and update

### LINKED ISSUE #34: Investigate and fix KILL command behavior
In #19 Steve was having troubles getting KILL to work with files that had spaces in the name. Based on the description of KILL [here](https://hwiegman.home.xs4all.nl/qb45-man/index.html) it seems like it should work if file names have spaces, but the C++ implementation `sub_kill()` looks fairly questionable and should be reviewed.

At the very least there is a clear difference between the Linux and Windows implementations because Windows uses `FindFirstFile()` which supposedly handles the `*` and `?` wildcards, but Linux just uses `remove()` which does not. Additionally, there is a lot of weird/suspicious path handling in Windows that might have been what was giving Steve some trouble. Certainly `DeleteFile()` itself can handle spaces in paths.

IMO it would probably be worth just implementing the `*` and `?` handling ourselves and use the same implementation for all platforms. That should hopefully also allow us to toss some of the weirder handling and get consistent behavior. We should also presumably verify the behavior of QuickBasic 4.5 a bit

### LINKED ISSUE #148: Refactor Generic File System API into source files in `libqb/`
The "Generic File System" API is a file access API that lives in `libqb.cpp`. It's purpose is to abstract away some of the more platform-specific details, and the API is fairly self-contained beyond `qbs` usage. We should look to refactor it into it's own source file in `libqb/`, along with a header for the API to be defined.

a740g:
Closes #34
Closes #148 

This pull request represents a significant refactoring effort to consolidate all filesystem-related APIs into their own translation unit inside `libqb/`. It now correctly handles all APIs, including some that were previously only implemented for Windows, across all supported platforms. This also auto-corrects directory separators in `func__loadimage()`, `sub__saveimage()`, `func__sndopen()` and `func__loadfont()`.

The following two new functions were added:
```vb
' MS BASIC PDS 7.1 & VBDOS 1.0 DIR$() style function 
_FILES$([fileSpec$])
```

```vb
' Returns the Fully Qualified Name for a file / directory
_FULLPATH$(pathName$)
```


SteveMcNeill:
I'm not so certain about the _FQN name.  Honestly, I don't think it's a set of abbreviations that's going to mean a whole lot to the old foggies who tend to program in BASIC a lot.  As much as I hate extra typing, I'd think a more verbose name might be needed here to maintain the general style as BASIC programming.  Perhaps something like _FULLPATH.  Or could these two functions somehow be combined into one?

_FILES$ (filespec$ , "verbose|brief")

Being picky over a command name seems almost petty in many ways, but this is very useful functionality.  You want to try and make it as intuitive for the general user as possible, when you add it into the language.

SteveMcNeill:
(Even if it is kept as a short command name, shouldn't it at least be _FQFN for a Fully Qualified File Name? )

a740g:
> I'm not so certain about the _FQN name. Honestly, I don't think it's a set of abbreviations that's going to mean a whole lot to the old foggies who tend to program in BASIC a lot. As much as I hate extra typing, I'd think a more verbose name might be needed here to maintain the general style as BASIC programming. Perhaps something like _FULLPATH. Or could these two functions somehow be combined into one?
> 
> _FILES$ (filespec$ , "verbose|brief")
> 
> Being picky over a command name seems almost petty in many ways, but this is very useful functionality. You want to try and make it as intuitive for the general user as possible, when you add it into the language.

BASIC itself is not consistent with names. 😁 FRE, MK* and CV* are some that comes to mind. But I too am with you on verbose names. I initially thought of `_FULLYQUALIFIEDNAME` but quickly dropped it after looking at the number of characters. 😊 _FULLPATH sounds good though. We should not combine `_FULLPATH` and `_FILES` because we should allow folks to get canonicalized absolute pathnames without going thought the internal setup that `_FILES` does.

I'll change `_FQN` to `_FULLPATH` unless someone has any other suggestions.

a740g:
> (Even if it is kept as a short command name, shouldn't it at least be _FQFN for a Fully Qualified File Name? )

Well, it works for both file names and directory names. So, I not sure we should use `_FQFN`. 🤷‍♂️

SteveMcNeill:
_FULLPATH sounds good to me,  unless anyone else has a better suggestion.  Otherwise, from what I can tell looking over this, it all looks good to me.  :)

SteveMcNeill (approved):

internal/c/libqb/src/filesystem.cpp
```diff
@@ -0,0 +1,986 @@
...
+    USER_FONTS,
+    TEMP,
+    PROGRAM_FILES,
+    PROGRAM_FILES_32,
+};
+
+/// @brief This populates path with the full path for a known directory
+/// @param kD Is a value from FS_KnownDirectory (above)
+/// @return Is the string that will receive the directory path. The string may be changed
+static std::string FS_GetKnownDirectory(FS_KnownDirectory kD) {
```
mkilgore, line 103: You might get something a bit cleaner if you have two separate `FS_GetKnownDirectory()` implementations for Windows Vs Unix. There would be some slight duplication in the structure, but a lot less `#ifdef`s to read.

mkilgore (approved):

internal/c/libqb/src/filesystem.cpp
a740g: That's a valid callout. It does look a little messy. I'll clean up that part.

a740g: This is done.

PR #427, opened 2024-01-04
clean up CONST somewhat.  Fix const names being used in other consts.

SteveMcNeill:
Believe it or not, CONST is still broke...  /sigh

I have no idea how nobody has complained yet, but currently CONST is not allowing us to use constants in our constants.

CONST foo = 1
CONST foo2 = foo + foo

The above for example.  ^^

Sure, we look up the const.  Sure we substitute it properly.  Sure, we do all we should do with it...

..and then we overwrite it and replace it back with our original string which we started with!   

[hr]

This should fix that.

This should also clean up some of the program flow and code logic to enable the no_prefix stuff to be easier to follow in the code, without backtracking and looping and using multiple loop values and everything that it's doing now.

### DIFF
--- a/source/qb64pe.bas
+++ b/source/qb64pe.bas
@@ -24509,6 +24509,7 @@ SUB Set_OrderOfOperations
     'I used a range here so I could add in new priority levels as needed.
     'OName ended up becoming the name of our commands, as I modified things.... Go figure!  LOL!
     REDIM OName(10000) AS STRING, PL(10000) AS INTEGER
+    SHARED Skip_NoPrefix_Low AS INTEGER, Skip_NoPrefix_Max AS INTEGER 'to skip the no_prefixed functions
     'Constants get evaluated first, with a Priority Level of 1
 
     i = i + 1: OName(i) = "C_UOF": PL(i) = 5 'convert to unsigned offset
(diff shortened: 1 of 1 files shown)


source/qb64pe.bas
```diff
@@ -24572,6 +24574,34 @@ SUB Set_OrderOfOperations
     i = i + 1: OName(i) = "_BLUE": PL(i) = 10
     i = i + 1: OName(i) = "_ALPHA": PL(i) = 10
 
+    Skip_NoPrefix_Low = i + 1
+    i = i + 1:: OName(i) = "PI": PL(i) = 10
```
mkilgore, line 24578: Is it really better to list all of these out separately rather than stripping the `_` during the check like we were before? It just seems like this is harder to maintain since we now have to make sure this list matches all the `_` entries listed above.

source/qb64pe.bas
```diff
@@ -24637,22 +24700,23 @@ FUNCTION EvaluateNumbers$ (p, num() AS STRING)
             'Note, these are special cases and work with the number BEFORE the command and not after
```
mkilgore, line 24700: If you look at line `24628`, I think the addition of the non-prefixed entries to `OName()` creates problems because situations like that `Select Case OName(p)` now need to be checking for both.

source/qb64pe.bas
```diff
@@ -24637,22 +24700,23 @@ FUNCTION EvaluateNumbers$ (p, num() AS STRING)
...
+                CASE "C_UBI": n1~` = VAL(num(1)): temp$ = STR$(n1~`)
+                CASE "C_BI": n1` = VAL(num(1)): temp$ = STR$(n1`)
+                CASE "C_FL": n1## = VAL(num(1)): temp$ = STR$(n1##)
+                CASE "C_DO": n1# = VAL(num(1)): temp$ = STR$(n1#)
+                CASE "C_SI": n1! = VAL(num(1)): temp$ = STR$(n1!)
             END SELECT
+            EvaluateNumbers$ = _TRIM$(temp$) + num(2)
             EXIT FUNCTION
         CASE 10 'functions
             SELECT CASE OName(p) 'Depending on our operator..
```
mkilgore, line 24722: Like the one I mentioned at `24682`, I think this select will have the same problem, it doesn't check for the non-underscore names.

source/qb64pe.bas
SteveMcNeill: I'd think so.  It should speed up the overall process of handling things, as it just skips these values if noprefix isn't in use, compared to previously where it ran those loops reagardless.  It's a little more code here, but less processing there.

source/qb64pe.bas
```diff
@@ -24607,6 +24636,39 @@ SUB Set_OrderOfOperations
...
+    PP_TypeMod(13) = "!": PP_ConvertedMod(13) = "C_SI" 'single
+    PP_TypeMod(14) = "##": PP_ConvertedMod(14) = "C_FL" 'float
+    PP_TypeMod(15) = "#": PP_ConvertedMod(15) = "C_DO" 'double
+    PP_TypeMod(16) = "_RGB32": PP_ConvertedMod(16) = "C_RG" 'rgb32
+    PP_TypeMod(17) = "_RGBA32": PP_ConvertedMod(17) = "C_RA" 'rgba32
+    PP_TypeMod(18) = "_RED32": PP_ConvertedMod(18) = "C_RX" 'red32
+    PP_TypeMod(19) = "_GREEN32": PP_ConvertedMod(19) = "C_GR" 'green32
+    PP_TypeMod(20) = "_BLUE32": PP_ConvertedMod(20) = "C_BL" 'blue32
+    PP_TypeMod(21) = "_ALPHA32": PP_ConvertedMod(21) = "C_AL" 'alpha32
+    PP_TypeMod(22) = "RGB32": PP_ConvertedMod(22) = "C_RG" 'rgb32
```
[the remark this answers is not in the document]
SteveMcNeill:
```
    FOR j = 1 TO UBOUND(PP_TypeMod)
        IF qb64prefix_set = 0 AND j > 21 THEN EXIT FOR 'only check the no_prefix junk if necessary
```

We only check those extra noprefix values if the flag is set.  Sure, the array might always hold those values, but we just exit the check once we get to them.

This seems more efficient and logical to me than the way it was storing two variables and then choosing which variable to use, which just obfuscates the code in that area making it more complex than it needs to be.

source/qb64pe.bas
SteveMcNeill: Good catch.  I meant to fix that and then just got myself sidetracked chasing down a different glitch.  These should now check both for us, properly.

source/qb64pe.bas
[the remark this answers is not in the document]
mkilgore: Oh I see - why are we hardcoding the `21` rather than place it in a variable like it was before?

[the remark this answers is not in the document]
SteveMcNeill: I can always make this a shared variable between those two subs, if you prefer, but it just makes more sense to me to have 1 variable for the cutoff point here, rather than 2 variables and then an if check to toggle between them as it was.  Heck even the following would make more sense and be easier to sort out: 

IF qb64prefext_set THEN j_limit = UBOUND(PP_TypeMod) ELSE j_limit = 21 'or a shared variable at that cutoff point
FOR j = 1 to j_limit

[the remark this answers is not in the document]
mkilgore: Yeah that part is fine, I just think the `21` shouldn't be hardcoded, we should set a variable for it back where the array is initialized.

source/qb64pe.bas
mkilgore: We're only talking about 20 or so strings, why is it worth it to make the code harder to maintain when the speed-up is presumably not noticeable?

I'm not so sure it's actually faster either, the `Select Case OName()`s now have to check for both the prefixed and non-prefixed versions and that's not conditional on `qb64prefix_set`.

SteveMcNeill:
I'll wait for Matt's changes and then work on cleaning things up from there a bit, if still needed.  ;)

PR #428, opened 2024-01-04
Fix OPEN COM from locking up inside gfs_open_com_syntax()

a740g:
This PR fixes an `OPEN COM` statement from locking up inside `gfs_open_com_syntax()`.

For example, the following code crashes with v3.9.0, v3.9.1 and locks up on later versions.

The issue was first reported here: https://qb64phoenix.com/forum/showthread.php?tid=2376&pid=22478#pid22478

```vb
ON ERROR GOTO Handler
OPEN "COM4:74880,N,8,1,BIN,CS0,DS0,RB8192" FOR RANDOM AS #99
END

Handler:
PRINT "Error:"; ERR
RESUME NEXT
```

With the changes in the PR, this no longer happens. The changes mostly consist of re-formatting the code and improving readability. But most importantly, it now returns `1` if it can completely parse the string.

### DIFF
--- a/internal/c/libqb.cpp
+++ b/internal/c/libqb.cpp
@@ -29174,7 +29174,7 @@ int32_t gfs_open_com_syntax(qbs *fstr, gfs_file_struct *f) {
     if (fstr->len <= 3 || (fstr->chr[0] & 223) != 67 || (fstr->chr[1] & 223) != 79 || (fstr->chr[2] & 223) != 77)
         return 0; // ! C/c,O/o,M/m
 
-    int32_t c, i, x, z, v = -1;
+    int32_t c, i, v = -1, x = 0;
     for (i = 3; i < fstr->len - 1; i++) {
         c = fstr->chr[i];
         if (c == ':')
@@ -29565,7 +29565,7 @@ comstatment:
 
     if (f->com_op_x == -1) {
         x = f->com_cd_x * 10;
-        z = f->com_ds_x * 10;
+        auto z = f->com_ds_x * 10;
         if (z > x)
             x = z;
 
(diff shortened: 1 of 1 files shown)


internal/c/libqb.cpp
```diff
@@ -29451,44 +29522,60 @@ int32 gfs_open_com_syntax(qbs *fstr, gfs_file_struct *f) {
...
+
     if (f->com_cs_x == -1)
         f->com_cs_x = 1000;
+
     if (f->com_ds_x == -1)
         f->com_ds_x = 1000;
+
     if (f->com_op_x == -1) {
         x = f->com_cd_x * 10;
         z = f->com_ds_x * 10;
```
mkilgore, line 29568: I think this is the only usage of `z` (hard to tell). If it is you might as well just declare it right here.

internal/c/libqb.cpp
```diff
@@ -29163,46 +29163,46 @@ int64 gfs_lof(int32 i) {
...
         }
-        v = v * 10 + (c - 48);
+        v = v * 10 + (c - '0');
     }
+
     return 0; // no ":"
+
 comstatment:
-    if ((x >= 7) || (v <= 0) || (v > 255))
+    if (x >= 7 || v <= 0 || v > 255)
```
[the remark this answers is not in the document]
a740g: Nice catch! I'll just initialize x to 0 here.

internal/c/libqb.cpp
a740g: Done.

mkilgore (approved):

PR #429, opened 2024-01-06
Fix using CONST values in CONST statements.

mkilgore:
This moves the CONST replacement up before we turn the elements into a single string. The advantage is that we don't have to worry about splitting the string properly to find the CONST names as the elements are already split for us.


a740g (approved):

SteveMcNeill (approved):

SteveMcNeill:
Question: 

```
                                        IF t AND ISSTRING THEN
                                            e2$ = conststring(i3)
```

On the segment above, does that e2$ need to be enclosed in quotes here?  If we're doing a simple substitution, it seems like we'd be missing them.

CONST myName$ = "Steve"

Does that store it in constring as "Steve" or just Steve?  I don't remember how we store that, and if it's the second, it could lead to issues rebuilding that string.

mkilgore:
@SteveMcNeill - It's a good question, it actually stores it in the intermediate `"foo",3` format that includes the quotes and string length. I didn't bother separating out just the quoted string as the next code (that handles the `&H` and friends) does that anyway.

RhoSigma-QB64 (approved):

PR #432, opened 2024-01-08
update from main

lenoil98:


SteveMcNeill:
Can we get some description of what this is, what it does, and why it should be added into the language.   It's always nice to have at least *some* type of idea of what one is going to be looking at and checking for, before trying to wade into changes blindly.  ;)

notpeter:
This appears to be a WIP PR which includes some modifications to get it to compile on FreeBSD. It's not suitable for merge (hard coded lib paths, readme edits, etc).

SteveMcNeill:
Closing this for now.

PR #444, opened 2024-02-02
Adler crc hash

RhoSigma-QB64:
This 1st PR will implement the new functions _ADLER32, _CRC32 and _MD5$, after it is successfully merged I'll start a 2nd PR which then exchanges the use of the recently added GetCRC32&() BASIC FUNCTION (utilities/strings.bas) with _CRC32 or probably even _MD5$ and remove that BASIC FUNCTION again.


internal/c/libqb/include/compression.h
```diff
@@ -9,5 +9,7 @@
 
 struct qbs;
 
+uint32_t func__adler32(qbs *text);
```
mkilgore, line 12: Rather than put them in `font.h` and `compression.h` I think it would make more sense to add a new `hashing.h` and put them in there.

mkilgore (approved):
Can you add some tests showing the new functionality working?

a740g:
> Can you add some tests showing the new functionality working?

I agree. A simple test with a minimum of 6 function calls in total should be conducted. For each of the 3 functions, we need two tests: one with an empty string and another with some pre-defined string.

Also like @mkilgore pointed, adler32 for an empty string should return a 1.

a740g (approved):

PR #454, opened 2024-02-16
Implement new file functions

RhoSigma-QB64:
- implements [_READFILE$](https://qb64phoenix.com/qb64wiki/index.php/READFILE$)
- implements [_WRITEFILE](https://qb64phoenix.com/qb64wiki/index.php/WRITEFILE)


internal/c/libqb.cpp
```diff
@@ -35923,6 +35923,48 @@ void sub__echo(qbs *message) {
...
+        fread(cont -> chr, 1, len, file);
+        fclose(file);
+        return cont;
+    } else {
+        error(QB_ERROR_FILE_NOT_FOUND); // error, we throw most common
+        return qbs_new_txt(""); // return empty on error
+    }
+}
+
+int16_t func__writefile(qbs *filespec, qbs *contents)
```
a740g, line 35949: Shouldn't this be a `SUB` instead of a `FUNCTION`?

We are already throwing an error. So, we can expect folks to trap the error if writing fails. Basically, either return a failure code or throw an error, not both.

I would side with throwing an error because it makes the usage simpler.

a740g (approved):

mkilgore (approved):

Issue #465, opened 2024-02-25
MOUSEMOVEMENTX/Y does not work properly on MacOS


TeddyKlugs:
The commands stop reporting position changes at the edges of the desktop instead of continuing infinitely. On my machine the extremes of the X movements are -50 to desktopwidth - 50. I haven't played around with Y movements.

a740g:
@TeddyKlugs Thank you.

Adding forum link where this was discussed and reported.
https://qb64phoenix.com/forum/showthread.php?tid=2445

SteveMcNeill:
Also want to add a note that _MouseMove is glitchy on Mac.  (At least IMHO, it is.)  Any time it's called, glut locks out all keyboard and mouse input for 250ms, for some odd reason.  This should also be looked into at the same time the _MouseMomentX/Y stuff is examined.  ;)

PR #468, opened 2024-03-09
Add support for _MOUSEWHEEL and _MOUSEMOVEMENTx on macOS

### LINKED ISSUE #289: `_MOUSEWHEEL` does not work on macOS
This has been a known limitation for awhile, it looks like there might be potential for fixing it by simply not using the GLUT provided by macOS. It will need some investigation.

a740g:
Fixes #289, #465

This PR fixes some longstanding macOS issues with the mouse.

Key fixes include:
- Enabling support for `_MOUSEWHEEL`.
- Implementing `_MOUSEMOVEMENTX` and `_MOUSEMOVEMENTY` accurately.

Previously, these issues stemmed from limitations within the GLUB library on macOS. To work around these limitations, this PR leverages the Quartz Event Services API in a manner that does not disrupt the GLUT input event queue, ensuring smooth and reliable mouse functionality.


internal/c/libqb/src/mac-mouse-support.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+    if (CGEventGetIntegerValueField(event, kCGEventTargetUnixProcessID) == appPID) {
+        if (type == kCGEventScrollWheel) {
+            auto deltaScroll = CGEventGetIntegerValueField(event, kCGScrollWheelEventPointDeltaAxis1);
+            GLUT_MOUSEWHEEL_FUNC(0, deltaScroll, g_MouseX, g_MouseY);
+        } else {
+            auto deltaX = CGEventGetIntegerValueField(event, kCGMouseEventDeltaX);
+            auto deltaY = CGEventGetIntegerValueField(event, kCGMouseEventDeltaY);
+            if (deltaX || deltaY)
+                qb64_custom_event(QB64_EVENT_RELATIVE_MOUSE_MOVEMENT, deltaX, deltaY, 0, 0, 0, 0, 0, 0, nullptr, nullptr);
```
mkilgore, line 36: I think it would be good to pull the `QB64_EVENT_RELATIVE_MOUSE_MOVEMENT` code out of `qb64_custom_event` and place it in a separate function in `libqb.cpp`. After that both `qb64_custom_event()` and your own code can call that new function.

`qb64_custom_event()` was designed so that there's only a single function for the GLUT code to call, but it's pretty nasty to use elsewhere with all the extra parameters and IMO worth avoiding.

internal/c/libqb/src/mac-mouse-support.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+extern "C" int qb64_custom_event(int event, int v1, int v2, int v3, int v4, int v5, int v6, int v7, int v8, void *p1, void *p2);
+void GLUT_MOUSEWHEEL_FUNC(int wheel, int direction, int x, int y);
+
+extern int g_MouseX, g_MouseY;
+
+static CFMachPortRef g_EventTap = nullptr;
+static CFRunLoopSourceRef g_RunLoopSource = nullptr;
+
+static CGEventRef macMouseCallback(CGEventTapProxy proxy, CGEventType type, CGEventRef event, void *userInfo) {
```
mkilgore, line 25: I assume the callback gets called from the same main thread that GLUT runs on (since that should be running the event loop)?

internal/c/libqb/src/glut-main-thread.cpp
```diff
@@ -178,8 +181,11 @@ void libqb_start_main_thread(int argc, char **argv) {
 // from two threads at the same time).
 //
 // This is accomplished by simply queuing a GLUT message that calls exit() for us.
-void libqb_exit(int exitcode)
-{
+void libqb_exit(int exitcode) {
+#ifdef QB64_MACOSX
+    macMouseDone();
```
mkilgore, line 186: I think this call belongs in [here](https://github.com/QB64-Phoenix-Edition/QB64pe/blob/c7e912ace636e90248710196e4b189af8dfc8967/internal/c/libqb/src/glut-message.cpp#L52) instead. I suspect that you're intended to only call some of those `CG*` functions from the main thread.

internal/c/libqb/src/mac-mouse-support.h
```diff
@@ -0,0 +1,6 @@
+#pragma once
+
+#ifdef QB64_MACOSX
```
mkilgore, line 3: It's a small thing in this case, but rather than require the caller to check `QB64_MACOSX` every time they want to call these functions I think it's better to declare stubs in here that don't do anything on the other OSs. That way callers don't have to check and we don't need to add `#ifdef`s everywhere. IE:

```c
#ifdef QB64_MACOSX
void macMouseInit();
void macMouseDone();
#else
static inline void macMouseInit() { }
static inline void macMouseDone() { }
#endif
```

mkilgore (approved):

internal/c/libqb/src/mac-mouse-support.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+#define QB64_EVENT_CLOSE 1
+#define QB64_EVENT_KEY 2
+#define QB64_EVENT_RELATIVE_MOUSE_MOVEMENT 3
+#define QB64_EVENT_FILE_DROP 4
+
+extern "C" int qb64_custom_event(int event, int v1, int v2, int v3, int v4, int v5, int v6, int v7, int v8, void *p1, void *p2);
+void GLUT_MOUSEWHEEL_FUNC(int wheel, int direction, int x, int y);
+
+extern int g_MouseX, g_MouseY;
```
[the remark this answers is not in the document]
a740g: Done.

internal/c/libqb/src/mac-mouse-support.cpp
a740g: Cleaned this up as suggested.

internal/c/libqb/src/glut-main-thread.cpp
a740g: Moved this to `glut_message_exit_program::execute()`.

internal/c/libqb/src/mac-mouse-support.h
a740g: Yeah. This looks a lot cleaner. Done.

internal/c/libqb/src/mac-mouse-support.cpp
a740g: Yes. None of this stuff depends on GLUT as it bypasses it and goes straight to the OS.

grymmjack:
Tested on my Mac:
gj-Mac-Specs

With this code:
```vb
DO
    ' Keyboard handling
    DIM k AS STRING
    DIM AS INTEGER MMX, MMY, MW
    k$ = LCASE$(INKEY$)
    DO WHILE _MOUSEINPUT
        MMX = _MOUSEMOVEMENTX
        MMY = _MOUSEMOVEMENTY
        MW = _MOUSEWHEEL
        IF MMX <> 0 THEN
            LOCATE 1, 1 : PRINT SPACE$(50)
            LOCATE 1, 1 : PRINT "MMX:", MMX
        END IF
        IF MMY <> 0 THEN
            LOCATE 2, 1 : PRINT SPACE$(50)
            LOCATE 2, 1 : PRINT "MMY:", MMY
        END IF
        IF MW <> 0 THEN
            LOCATE 3, 1 : PRINT SPACE$(50)
            LOCATE 3, 1 : PRINT "MW:", MW
        END IF
    LOOP
    IF k$ = CHR$(27) THEN EXIT DO
LOOP
SYSTEM
```

Worked great!

image

built from the PR branch `a740g:macOS-mouse-fixes`

RhoSigma-QB64 (approved):

Issue #408, opened 2023-11-20
The GLUT thread does not accurately redraw at 60 FPS


mkilgore:
The GLUT thread does not accurately do the redrawing at 60 FPS as intended, instead it's a bit slow. It's hard to see this in action as the GLUT thread runs independent of the rest of the QB64 code, but if you measure the FPS directly from the GLUT thread it's very obvious.

The issue is simple, the [GLUT timer func](https://github.com/QB64-Phoenix-Edition/QB64pe/blob/e634c4b028ff8467655eaff6fdf43ed38bb11b3e/internal/c/libqb.cpp#L34139) (or idle func on Linux and Mac OS) does not properly calculate the time till the next frame. The math simply calculates how much time is in-between each frame and fails to account for the time it took to render the last frame, causing it to be slightly slow. To get it really accurate, we also need to account for the error introduced from the `Sleep()` that we do as it does not always last exactly the number of milliseconds we give it.

A rough fixed version for Windows is below. Basically, it uses the time to measure how long the last frame took to render, and then sleeps for the remaining time in the frame. It also measure how long the sleep actually took and compensates for that in the next frame:

```c
static int64_t lastTick = 0;
static double deltaTick = 0;

void GLUT_IDLEFUNC() {
    libqb_process_glut_queue();

#    ifdef QB64_GLUT
    int64_t curTime = GetTicks();

    // This is how long the frame took to render
    int64_t elapsed = curTime - lastTick;

    // Calculate out the error between how long the frame was 'supposed' to take vs. how long it actually took.
    deltaTick += ((double)1000 / max_fps) - (double)elapsed;

    lastTick = curTime;

    // If the error is positive, we sleep for that period of time.
    if (deltaTick > 0) {
        int32 msdelay = deltaTick;
        Sleep(msdelay);

        int64_t sleepTime = GetTicks();

        // Subtract off the time we spent sleeping. This should leave deltaTick at zero or slightly negative.
        // If it ends up negative, then we'll sleep less next frame to compensate
        deltaTick -= sleepTime - lastTick;
        lastTick = sleepTime;
    }

    glutPostRedisplay();
#    endif
}
```

The Mac OS and Linux ones would be basically the same, we could probably simplify it all into one function.

Issue #409, opened 2023-11-20
Provide a way to wait for the current queued frame to be rendered on screen.


mkilgore:
Currently there is no way to sync `_Display` requests with the GLUT thread that renders the frames in the window. The result is that even if you use `_Limit 60` and the GLUT thread runs at 60 FPS, occasionally things won't line up and you'll miss a frame or get a frame dropped.

I think the easiest way to do it would be to combine the waiting with `_Display` and have a `_Display _Wait`. We'd have to decide on the behavior, but I think the most obvious is to have it render the frame like normal and then wait for that frame to be picked up by the GLUT thread. With that behavior, the QB64 program would then have a bit less than 1/60 of a second to do its work and call `_Display _Wait` again to ensure a smooth 60 FPS.

PR #474, opened 2024-03-25
Fix GLUT thread to redraw at an accurate 60 FPS.

mkilgore:
Currently the GLUT thread draws the screen too slowly, it is supposed to
default to 60 FPS but you'll always get a bit less than that. This is because
the current logic simply inserts delays for the entire length of a frame, not
taking into account how long it took us to render the last frame. Any time
spent rendering thus results in a slowed down FPS.

The new code uses GetTicks() to measure how much time has passed since
the last render, which then lets us calculate the exact amount of time
until the next frame so we can sleep for that amount. We additionally
then measure how long the sleep lasted vs. what we asked for (since any
sleep we do only has a minimum guarantee, it will occasionally last a bit
longer) and adjust based on that as well. The result is a perfect 60 FPS
as long as rendering is quick enough.

If the rendering falls behind (Ex. a slow _GL SUB is in the program)
then we'll start skipping frames to get back on track. This is behavior
may want to be talk about  or perhaps make configurable, I went with
basically having it drop back to 30 FPS (and then 20/15/etc.) if the rendering
is too slow. The alternative would be that we just set the `deltaTick` back to zero
and render as fast as possible.

The undocumented `_FPS` command can be used to change the speed from
60 FPS to whatever is desired (with a cap of 200 FPS). This was always
true, but now it should be accurate to whatever you request.

Fixes: #408

### DIFF
--- a/internal/c/libqb.cpp
+++ b/internal/c/libqb.cpp
@@ -29468,6 +29468,8 @@ int32 func__scaledheight() { return environment_2d__screen_scaled_height; }
 extern void set_dynamic_info();
 
 int main(int argc, char *argv[]) {
+    clock_init();
+
 
 #if defined(QB64_LINUX) && defined(X11)
     XInitThreads();
--- a/internal/c/libqb.cpp
+++ b/internal/c/libqb.cpp
@@ -27227,49 +27227,61 @@ void GLUT_SPECIAL_FUNC(int key, int x, int y) {
 }
 void GLUT_SPECIALUP_FUNC(int key, int x, int y) { GLUT_key_special(key, 0); }
 
-#ifdef QB64_WINDOWS
-void GLUT_TIMER_EVENT(int ignore) {
-    libqb_process_glut_queue();
+static int64_t lastTick = 0;
+static double deltaTick = 0;
 
-#    ifdef QB64_GLUT
-    glutPostRedisplay();
-    int32 msdelay = 1000.0 / max_fps;
-    Sleep(4);
-    msdelay -= 4; // this forces GLUT to relinquish some CPU time to other threads but still allow for _FPS 100+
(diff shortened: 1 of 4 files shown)


a740g (approved):
@mkilgore Do you think we should document the _FPS comamnd?

RhoSigma-QB64 (approved):

mkilgore:
@a740g Maybe, but the `_Auto` flag is not implemented so that might be a bit of a problem. Based on it's description of "make QB64 auto-adjust fps based on load" I'd rather just get rid of it, functionally we already do that - if the CPU is too slow then the FPS goes down :rofl: 

What would probably be a good idea is replacing `_Auto` with a setting to toggle between preferring a stable FPS vs. preferring as many frames as possible under the requested FPS, but I don't really want to mess with implementing that at the moment.

grymmjack:
@mkilgore will this fix the stuttering that I shared here: 
https://app.screencast.com/7XGWZY42TJJcE

Thank you for this contribution regardless! <3

a740g:
> @mkilgore will this fix the stuttering that I shared here: https://app.screencast.com/7XGWZY42TJJcE
> 
> Thank you for this contribution regardless! <3

I think it should.

Thank you @mkilgore for fixing this! You should really push the merge button now. 😁

mkilgore:
> @mkilgore will this fix the stuttering that I shared here: https://app.screencast.com/7XGWZY42TJJcE
> 
> Thank you for this contribution regardless! <3

Like @a740g said I think it should probably fix some of it, you should see it move at 60 FPS after this. Some of the stutter will be due to syncing between the QB64 thread and the rendering thread and that's not addressed by this. #409 would be a complete fix for that but will also require some level of code modification to use.

PR #476, opened 2024-03-30
Avoid monitoring special keys when the window is not in focus

a740g:
This PR fixes the issue reported [here](https://qb64phoenix.com/forum/showthread.php?tid=2504&pid=23718#pid23718).

Instead of making custom modifications to FreeGLUT, we will simply add a check within the `GLUT_key_special()` function to monitor special keys only when the QB64 window is in focus.

### DIFF
--- a/internal/c/libqb.cpp
+++ b/internal/c/libqb.cpp
@@ -27544,34 +27544,28 @@ void GLUT_KEYBOARDUP_FUNC(unsigned char key, int x, int y) { GLUT_key_ascii(key,
 
 void GLUT_key_special(int32 key, int32 down) {
 #ifdef QB64_GLUT
-
-#    ifdef QB64_WINDOWS
-    if (!func__hasfocus())
-        return;
-#    endif
-
 #    ifndef CORE_FREEGLUT
-        /*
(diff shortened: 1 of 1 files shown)


RhoSigma-QB64 (approved):

internal/c/libqb.cpp
```diff
@@ -27544,28 +27544,34 @@ void GLUT_KEYBOARDUP_FUNC(unsigned char key, int x, int y) { GLUT_key_ascii(key,
 
 void GLUT_key_special(int32 key, int32 down) {
 #ifdef QB64_GLUT
+
+#    ifdef QB64_WINDOWS
+    if (!func__hasfocus())
```
mkilgore, line 27549: Previous versions will still report the release even if focus is lost while the key is held. I think we can match that if we move this past the `vk = ` conversion logic and then do `if (!func__hasfocus() && !(keyheld(vk) && !down))`

mkilgore (approved):

internal/c/libqb.cpp
a740g: Made changes as suggested.

PR #478, opened 2024-04-06
Logical Boolean operators

### LINKED ISSUE #211: Boolean short-circuiting versions of `AND` and `OR`
We need Boolean short-circuit evaluation versions of `AND` and `OR`.
VB, FreeBASIC and many other languages use the `AndAlso` and `OrElse` keywords.
This can be implemented in QB64 using `_ANDALSO` and `_ORELSE` keywords.

```VB
IF x = 0 _ORELSE y = 0 THEN
...
END IF
```

```VB
IF x = 0 _ANDALSO y = 0 THEN
...
END IF
```

@mkilgore also noted that `AND` and `OR` in QB64 are not logical operators. So, `_ANDALSO` and `_ORELSE` keywords can be implemented as logical Boolean short-circuit evaluation versions of `AND` and `OR`. Apparently, `NOT` is also not logical which can lead to surprising situations (ex. `NOT 1` does not give a `false` value). So, we may need a logical version of `NOT` as well (ex. `_LNOT`)

See [Logical and Bitwise Operators in Visual Basic](https://learn.microsoft.com/en-us/dotnet/visual-basic/programming-guide/language-features/operators-and-expressions/logical-and-bitwise-operators) for a more in-depth discussion on logical and bitwise operators.

a740g:
Closes #211

This PR implements three Boolean logical operators:

1. **_NEGATE**: A unary logical operator that performs logical negation on a Boolean expression.
2. **_ANDALSO**: A binary logical operator that performs short-circuiting logical conjunction on two Boolean expressions.
3. **_ORELSE**: A binary logical operator that performs short-circuiting logical inclusion on two Boolean expressions.

A test is included to ensure we get the expected behavior.


RhoSigma-QB64 (approved):
Works as advertised :)

PR #490, opened 2024-05-09
Refactor RAD v2 Opal OPL3 FM Emulator

a740g:
This PR refactors the RAD v2 Opal OPL3 FM emulator by splitting it into separate .h and .cpp files. The goal is to enable easy future reuse.

Currently, the emulator is utilized for playing RAD v1 and v2 files. However, it is versatile enough to serve as a basic MIDI back-end.

By using it as a MIDI back-end, one potential benefit will be the reduced executable size, achieved by avoiding the inclusion of the 1 MB soundfont payload in the final executable. We can also simplify the makefile logic by eliminating the conditional soundfont and player code inclusion.


RhoSigma-QB64 (approved):

grymmjack:
How would we use this FM a for MIDI ? Do we have GM spec sounds?

a740g:
> How would we use this FM for MIDI ? Do we have GM spec sounds? Like E-Piano or such?

Something like https://github.com/devinacker/ymfmidi. I already got something working using Nuked-OPL3 and ymfm. So, reworking that for Opal should be trivial. For MIDI patches we can use https://github.com/sneakernets/DMXOPL. My idea is to use a generic integrated default patch and then let users load custom patches or a soundfont (if they want to use the soundfont based player).

grymmjack:
Oh! I remember us discussing DMXOPL before now that you mention it having looked at the site and jogging my memory. I think I found a VST to use these patches? Man it wasn't that long ago, but my memory ... 

@a740g sweet! I'd love to see/play/test things if you want a tester/nerd friend on any of this music stuff.

PR #491, opened 2024-05-12
Multiple font related optimization and improvements

a740g:
### Changes in this PR:

- The automatic fixed-width font detection feature is now disabled by default. Users can enable it by passing the "AUTOMONO" flag. Additionally, font flag changes are now correctly passed upstream to `libqb`.

- Font baseline calculation no longer relies on floating-point math. Instead, we utilize `FreeType` fixed-point functions such as `FT_MulDiv()` and `FT_MulFix()`. The baseline calculation method has also been refined for greater accuracy, ensuring precise baseline values.

- Significant enhancements have been made to the `_U*` functions. They now give more precise baseline-to-baseline height and font height measurements for both built-in and `FreeType` supported fonts. Superfluous adjustments that were previously applied have been eliminated, and all calculations now avoid floating-point math.

- Various quality-of-life enhancements have been implemented, including the usage of defined values instead of numeric literals and the use of `std::string` instead of `qbs` in `func__loadfont()`.


RhoSigma-QB64 (approved):
Works for me, tested my game with various fonts (fixed and proportional) and all looks as expected again.

SteveMcNeill (approved):
From what little testing I've did of these, they appear to be all working as advertised.  Many thanks for keeping on these as much as you have.  I appreciate it, and I know the vast majority of our user baser does too.  :)

PR #510, opened 2024-06-22
Improve func__deflate() & func__inflate()

a740g:
This PR does the following:

- Optimizes `func__deflate()` & `func__inflate()` to eliminate unnecessary buffer copies. This results in a nice speed boost of upto 15% in some cases on my Ryzen 5600
- Removes all unsafe usage of `malloc()` and `free()`
- Handles a case where the user can pass negative values for `originalSize` which can potentially break `qbs`
- Delegates the linking responsibility to the linker. This makes it possible to use the library from elsewhere with ease [needed for an upcoming PR]

### DIFF
--- a/internal/c/parts/compression/compression.cpp
+++ b/internal/c/parts/compression/compression.cpp
@@ -7,52 +7,54 @@
 
 #include "compression.h"
 #include "qbs.h"
+
 #include "miniz.h"
 
+#include <vector>
+
 uint32_t func__adler32(qbs *text) {
-    if (!text->len) return 1;
-    return (uint32_t) adler32(1, text->chr, text->len);
+    if (!text->len)
+        return 1;
+    return (uint32_t)adler32(1, text->chr, text->len);
 }
 
 uint32_t func__crc32(qbs *text) {
-    if (!text->len) return 0;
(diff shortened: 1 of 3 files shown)


internal/c/parts/compression/compression.cpp
```diff
@@ -7,52 +7,54 @@
...
-    int32_t result = compress(dest, &compsize, text->chr, filesize);
-    qbs *ret = qbs_new(compsize, 1);
-    memcpy(ret->chr, dest, compsize);
-    free(dest);
-    return ret;
+    auto fileSize = uLongf(text->len);
+    auto compSize = compressBound(fileSize);
+    auto dest = qbs_new(compSize, 1);                    // compressing directly to the qbs gives us a performance boost
+    compress(dest->chr, &compSize, text->chr, fileSize); // discard result because we do not do any error checking
+    qbs_left(dest, compSize);                            // discard the return value because it will return the same tmp qbs anyway
```
flukiluke, line 32: Just write `return qbs_left(dest, compSize)`, you get to save a line and it avoids relying on `qbs_left` to perform a particular kind of optimisation

a740g: Done.

flukiluke (approved):

PR #518, opened 2024-07-10
Move ide files

RhoSigma-QB64:
- move IDE related config and workfiles out of `internal/temp` into `settings` directly under the `qb64pe` folder
  - note this folder is not maintained in the repo, but created on the first IDE run, the user has a choice to copy his settings from another QB64-PE installation or to continue with default settings
- the *recent files* and *search string* histories are globally shared now by all running IDE instances
  - if one instance makes changes to such a history, then all other running instances will be sync'ed, the new/changed entries will be immediately available in all instances
- bookmarks behavior remains unchanged, those are saved/restored globally connected to the respective basic source files
- undo behavior remains unchanged and is managed per instance
---
- as a side change I've added 4 new preset color schemes, *Cornfield*, *Broadcast*, *X11 SgiColors* by me and *VS Code* curtesy to @a740g


source/ide/ide_histories.bas
```diff
@@ -0,0 +1,172 @@
...
+        IF UCASE$(be$) = ue$ OR lc% = 100 THEN 'already known or limit reached?
+            nul& = SeekBuf&(bh%, -LEN(BufEolSeq$(bh%)), SBM_BufCurrent) 'back to prev (just read) line
+            nul& = SeekBuf&(bh%, 0, SBM_LineStart) 'and then ahead to the start of it
+            DeleteBufLine bh%
+            EXIT WHILE
+        END IF
+    WEND
+    nul& = SeekBuf&(bh%, 0, SBM_BufStart) 'rewind
+    WriteBufLine bh%, e$ 'put new (or known) in 1st position (again)
+    UpdateHistory which$
```
mkilgore, line 25: I would think we want all of `AddToHistory` to happen while the lock is acquired. We could read/sync, modify. and write back all while holding the lock, ensuring that the in-memory version isn't out-of-date with the file when we make the modification.

source/ide/ide_filelock.h
```diff
@@ -0,0 +1,65 @@
...
+
+// Try to lock the given file to prevent other IDE instances from
+// accessing the same file at the same time.
+//  In: filename (STRING, add CHR$(0) to end of string)
+// Out: error    (INTEGER, 0 = no error (file locked), -1 = error (file not locked))
+//--------------------------------------------------------------------
+int32_t LockGlobalFileAccess(const char *filename) {
+    int32_t err = -1;     // so far assume error
+    if (filedesc == -1) { // make sure only one file at a time
+        filedesc = open(filename, O_RDWR);
```
mkilgore, line 39: I think we should differentiate the `open()` failing and the `flock()` failing. If the `open()` fails due to something like a permissions issue then we'll want to tell the user about that (`UpdateHistory()` would get stuck in a loop).

source/ide/ide_methods.bas
```diff
@@ -1312,6 +1306,8 @@ FUNCTION ide2 (ignore)
         exitvalue = _EXIT
         IF (exitvalue AND 1) <> 0 OR ideexit <> 0 THEN ideexit = 0: GOTO quickexit
 
+        IF _FILEEXISTS(RecentFile$ + FlagExt$) THEN SyncHistory "RECENT" 'recent file list changed, sync required
```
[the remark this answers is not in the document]
RhoSigma-QB64: Flag/signal file existence just triggers the sync, it's content otherwise is used by instances to  login/logout themself form sync, so the file is required.

source/ide/ide_histories.bas
RhoSigma-QB64: AddToHistory itself is a buffer operation, UpdateHistory is called to do the locked buffer2file update.

source/ide/ide_filelock.h
RhoSigma-QB64: I'll chack that tomorrow.

source/ide/ide_methods.bas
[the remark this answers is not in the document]
mkilgore: The flag file is only necessary to signal that the recent or history file has changed, correct? The mtime can tell you that without having to continually open/read the flag file, so it's significantly cheaper.

I guess separately, I'd be concerned that the instUseCnt will end up wrong when someone kills QB64 via TaskManager (or it crashes) rather than a clean exit.

source/ide/ide_histories.bas
mkilgore: I get the thinking, the issue I see is that if two instances end up in `AddToHIstory` at the same time then one of the history changes will be lost. If the lock is acquire before updating the in-memory copy then that issue can't happen.

RhoSigma-QB64: As each and every history change is directly connected to any kind of user interaction such as open a file, enter a search word or select "Clean history" from the menu, there can't be 2 or more instances fighting against each other. Even if you prepare everything and perfectly arrange the IDE windows so you just need to click to the next window and press enter to trigger such an action, I doubt you're faster then update process.

source/ide/ide_methods.bas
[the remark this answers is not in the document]
mkilgore: This is in the main IDE loop, isn't it? If so, then the IDE will open and read that sync file ~100 times a second, that's my concern.

source/ide/ide_filelock.h
RhoSigma-QB64: Well this one...
It's a use case scenario here, no intention to create something commonly usable. Thinking about the if and when those functions are called (and I always do such logic assumptions when coding, case it can save a lot of work) I see no reason why the open should fail at all, so to say I could just remove the open() fail check completely. The path alraedy exists and must be accessible, otherwise the none of the files could be read. The file I try to lock does already exist (the signal file), so we can assume if something would prevent accessing/writing the file that would already happen long before trying to open/lock the file. All we could do is to catch it and telling the user to check and possibly whitelist his qb64pe folder, but as we already do mention that in the Wiki, Forum, ReadMe... I don't currently see a need to add another such mention in code.

source/ide/ide_methods.bas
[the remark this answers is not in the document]
RhoSigma-QB64: Do you really think I wouldn't have checked for that impact before implementing it that way. A loop of 100000 OPEN/GET/PUT/CLOSE operations took ~13 seconds on my 14 years old laptop with mechanical harddrive. That's means an average of somewhat more than a 100th second for that 100 loops.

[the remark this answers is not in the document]
mkilgore: If you check the `mtime` then the whole issue goes away, you don't have to have that risk of slowdown or have to deal with the flag file at all.

Based on your comment about `AddToHistory` you probably wouldn't need the locking either, since two instances will never be writing to the history file at the same time.

source/ide/ide_histories.bas
```diff
@@ -0,0 +1,172 @@
...
+        result = idemessagebox("Remove Broken Links", "All files in the list are accessible.", "#OK")
+    ELSE
+        UpdateHistory "RECENT"
+    END IF
+END SUB
+
+'Load the search history into the specified array used by the IDE.
+'---------------------------------------------------------------------
+SUB RetrieveSearchHistory (SearchHistory() AS STRING)
+    bh% = OpenBuffer%("I", SearchedFile$) 'load and/or set pos (back) to start
```
mkilgore, line 160: Should you not be taking the lock here? If the buffer was recently cleared (or this is the first time reading it) then there doesn't seen to be anything to prevent the `OpenBuffer` call from reading the history file while another instance is in the middle of updating it.

PR #527, opened 2024-08-07
Udt string offsets & bump to version 3.14.0

### LINKED ISSUE #524: Variable length strings in deeply nested UDTs are not initialised
The following program crashes with a segmentation fault on the last line:
```
Type t1
    a As String
End Type

Type t2
    b As t1
    s As String
End Type

Type t3
    n As Long
    c As t2
End Type

Dim test As t3

Print test.c.s 'OK
Print test.c.b.a 'Segmentation fault
```

Initially observed at https://qb64phoenix.com/forum/showthread.php?tid=2889

Reproduced on Linux, commit 35d901fa6. At a glance the bad pointer looks to be the high 4 bytes of a valid pointer shifted down, so I suspect a misaligned store in the initialisation.

flukiluke:
Fixes an error in the initialisation of variable-length strings in UDTs. Fixes #524 

Also bumps the version to 3.14.0, on the assumption that this is the last feature to go in before release.

### DIFF
--- a/source/global/version.bas
+++ b/source/global/version.bas
@@ -1,9 +1,9 @@
 DIM SHARED Version AS STRING
 DIM SHARED IsCiVersion AS _BYTE
 
-Version$ = "3.13.1"
-$VERSIONINFO:FILEVERSION#=3,13,1,0
-$VERSIONINFO:PRODUCTVERSION#=3,13,1,0
+Version$ = "3.14.0"
+$VERSIONINFO:FILEVERSION#=3,14,0,0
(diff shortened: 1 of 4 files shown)


a740g (approved):

RhoSigma-QB64 (approved):

grymmjack (approved):

FellippeHeitor:
Will variable watch/set be affected?

flukiluke:
No, variable watching and setting is unaffected.

PR #530, opened 2024-08-19
Fixes and enhancements

RhoSigma-QB64:
Look into the commit comments.


internal/c/libqb/src/error_handle.cpp
```diff
@@ -16,6 +16,7 @@ uint32_t error_occurred;
 uint32_t error_retry;
 uint32_t error_err; //=0;
 uint32_t error_goto_line;
+qbs *error_handler_history;
```
mkilgore, line 19: Rather than a `qbs` could you use a `std::stack<uint32_t>`? It seems like that's what you're created and `std::stack` would avoid all the string manipulation.

PR #539, opened 2024-09-14
Update setup_mingw.cmd to use LLVM MinGW on all versions of Windows

a740g:
The PR makes the LLVM MinGW toolchain the only one used on Windows x86, x86-64, AArch32, and AArch64.

### DIFF
--- a/source/qb64pe.bas
+++ b/source/qb64pe.bas
@@ -12595,14 +12595,21 @@ IF os$ = "WIN" THEN
             IF n > 1 THEN a$ = "Unable to resolve multiple instances of sub/function '" + ResolveStaticFunction_Name(x) + "' in '" + ResolveStaticFunction_File(x) + "'": GOTO errmes
 
             IF n = 0 THEN 'attempt to locate simple function name without brackets
-                s$ = " " + ResolveStaticFunction_Name(x)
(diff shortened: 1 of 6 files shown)


source/qb64pe.bas
```diff
@@ -12601,8 +12601,8 @@ IF os$ = "WIN" THEN
                     a$ = ReadBufLine$(fh)
```
RhoSigma-QB64, line 12601: Ok, before messing with line endings I would suggest:
1.) revert this + CHR(10) changes you did, as ReadBufLine$() does stop at EOLs, we rather should tell the buffer to work with Linux/Mac EOLs instead
2.) to do so I think it's time now to replace the reduced utilities/s-buffer version with the full fledged version, grab the simplebuffer.bi/.bm files here https://github.com/RhoSigma-QB64/QB64Library/tree/main/QB64Library/SB-Storage and drop it in the utilities/s-buffer folder just overwriting the old ones
3.) Now place a `ConvBufToLnxMacEol fh` line right after the OpenBuffer%() call above and after the OpenBuffer%() in the second block below

a740g: Done. 👍

RhoSigma-QB64 (approved):

PR #543, opened 2024-09-22
Silence Linux C++ compiler warnings

a740g:
The PR silences many compiler warnings generated when compiling under Linux using gcc.

### DIFF
--- a/internal/c/libqb.cpp
+++ b/internal/c/libqb.cpp
@@ -1854,46 +1854,6 @@ void GLUT_DISPLAY_REQUEST();
 
 extern qbs *WHATISMYIP();
 
-// directory access defines
-#define EPERM 1
-#define ENOENT 2
-#define ESRCH 3
-#define EINTR 4
-#define EIO 5
-#define ENXIO 6
-#define E2BIG 7
-#define ENOEXEC 8
-#define EBADF 9
-#define ECHILD 10
-#define EAGAIN 11
-#define ENOMEM 12
-#define EACCES 13
-#define EFAULT 14
-#define EBUSY 16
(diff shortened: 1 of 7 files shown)


internal/c/libqb.cpp
```diff
@@ -13192,17 +13152,9 @@ void qbs_input(int32 numvariables, uint8 newline) {
                 c->chr[0] = 0;
                 qbs_set(qbs_input_arguements[argn], qbs_add(qbs_input_arguements[argn], c));
                 if (qbs_input_variabletypes[argn] & ISUNSIGNED) {
-#ifdef QB64_WINDOWS
-                    sscanf((char *)qbs_input_arguements[argn]->chr, "%I64u", (uint64 *)qbs_input_variableoffsets[argn]);
-#else
-                    sscanf((char *)qbs_input_arguements[argn]->chr, "%llu", (uint64 *)qbs_input_variableoffsets[argn]);
-#endif
+                    sscanf((char *)qbs_input_arguements[argn]->chr, "%" PRIu64, (uint64_t *)qbs_input_variableoffsets[argn]);
                 } else {
```
RhoSigma-QB64, line 13156: Shouldn't the respective SCN... macros be used here (lines above and below) for scanf(), instead of the PRI... ones supposed for printf() just for clarity https://pubs.opengroup.org/onlinepubs/009695399/basedefs/inttypes.h.html

RhoSigma-QB64 (approved):

internal/c/libqb.cpp
a740g: Good catch.

PR #562, opened 2024-11-03
Various refactoring

### LINKED ISSUE #29: Need a preprocessor identifier when $DEBUG is used
Not sure if we already have this. But I checked the wiki and found nothing.

Having an identifier (like DEBUG for example) auto-set when the $DEBUG metacommand is used can be useful to include code that we only need when debugging.

For example:

$DEBUG

$If DEBUG Then
             Print "We are debugging"
$End If

RhoSigma-QB64:
This PR removes the unstable state from the HTTP support. The other big part is refactoring several parts to benefit from the shortly added auto-includes. It also adds a pre-compiler flag indicating the $DEBUG state, this closes #29 .

It's recommended to review commit by commit, the commit notes give more informations.


internal/support/color/color32.bi
```diff
@@ -274,3 +274,26 @@ CONST Yellow~& = 4294967040
 CONST YellowGreen~& = 4288335154
 CONST YellowOrange~& = 4294946370
 
+'Color constants for text mode.
+'Just in case someone really needs both color sets in the same program.
+'These are equal to $COLOR:0 constants, but prefixed with "z" (for zero)
+'to avoid conflicts with same named constants in the block above.
+
+CONST zBlack~%% = 0
```
mkilgore, line 282: I don't think it's safe to add more of these since they lack the underscore :-/

mkilgore (approved):

PR #566, opened 2024-11-10
Add built-in logging support and other fixes

mkilgore:
This is getting close to being ready so I wanted to get some eyes on it:

This adds new logging capabilities to QB64, both internally from `libqb` and also exposed to the QB64 code. Beyond being a bit nicer than `Print`s since it captures the function name and line number, a big feature is that logging an error captures a stacktrace of the program at that point. Logging can also easily be turned on without recompilation via environment variables, and can be filtered via various scopes and a couple different logging levels. Overall this should hopefully be a big aid for debugging, as we can ask users to provide us with the logging output of programs when they error.

Below is an example of some logging from a QB64 program (note how the IDE opens a separate console window to display it):
image

The new commands are `_LogTrace`, `_LogInfo`, `_LogWarn`, `_LogError`, `_LogMinLevel`. The first four write out log messages at the indicated log level. The last returns a number indicating the lowest level of logging is enabled, allowing you to skip generating more expensive logging if it won't be displayed.

Logging is enabled via a variety of different environment variables (when run from the IDE these are configured automatically):
1. `QB64PE_LOG_LEVEL` - sets the lowest log level to display, options are `Trace`, `Information`, `Warning`, `Error`
2. `QB64PE_LOG_SCOPES` - A comma-separated list of different sets of logging to turn on, allowing basic filtering of what is displayed. Options are:
  a. `libqb` - Some generic logging from `libqb`, such as error codes when they happen, program startup/shutdown, etc.
  b. `libqb-image` - Logging from the image subsystem (when images are loaded,freed, errors that happen, etc.)
  c. `libqb-audio` - Logging from the audio subsystem
  d. `qb64` - Logging from the QB64 Program ( `_Log*` statements)
3. `QB64PE_LOG_HANDLERS` - A comma-separated list of the logging handlers to enable. If no handlers are enabled, then no logging is generated. Current options are:
  a. `console` - Logging output is written to the console.
  b. `file` - Writes logs to the file set via the `QB64PE_LOG_FILE_PATH` environment variable

Currently the IDE just defaults to using `Information` and enabling all the scopes, in a future update I'd like to have an IDE dialog that lets you pick the log level and scopes.


a740g (approved):

mkilgore:
This is ready to be reviewed, there's a couple other things to do (IDE dialog for configuring logging, convert the audio logging, coloring the output, etc.) but those can be left for separate PRs.

RhoSigma-QB64 (approved):

PR #568, opened 2024-11-18
miniaudio file name hashing fix

a740g:
This PR works around `miniaudio`'s file name hashing which causes the following issues:

1. When multiple files are extracted (e.g., from a game data archive) into the same temporary file (e.g., `temp.wav`) and loaded sequentially, all files are resolved to the first-loaded sound due to the static file name.

2. The MIDI file decoder allows user-defined SoundFonts (`.sf2`). However, reloading the same MIDI file with a different SoundFont causes a conflict due to file name hashing. For example: If MIDI file `A.mid` is loaded with SoundFont `A.sf2` and later with `B.sf2`, playback defaults to the first-loaded SoundFont (`A.sf2`), even if `B.sf2` is specified.

Additionally, `_MEMSOUND` now works correctly with fully decoded `MEMORY` sounds, expanding its functionality.

Quality of life improvements:

1. Removed usage of unsafe C macros.
2. Eliminated reliance on ma_decoder to simplify audio library.

Special thanks to @mkilgore for proposing the workaround and providing the original implementation.


internal/c/parts/audio/audio.cpp
```diff
@@ -2654,21 +3013,50 @@ int32_t func__sndopen(qbs *qbsFileName, qbs *qbsRequirements, int32_t passed) {
...
+                audioEngine.soundHandles[handle]->isUsed = false;
+
+                return AudioEngine::INVALID_SOUND_HANDLE;
+            }
+
+            AUDIO_DEBUG_PRINT("Sound length: %zd", contents.length());
+
+            // Configure a miniaudio decoder to load the sound from memory
+            audioEngine.soundHandles[handle]->bufferKey = std::hash<std::string_view>{}(contents);                            // make a unique key and save it
+            audioEngine.bufferMap.AddBuffer(contents.data(), contents.length(), audioEngine.soundHandles[handle]->bufferKey); // make a copy of the buffer
```
mkilgore, line 3045: IMO it would be worth avoiding the copy of the file contents when putting the buffer into the BufferMap, that was something I disliked about my simple solution.

mkilgore (approved):
Thanks for turning this into a full-fledged solution :)

RhoSigma-QB64 (approved):

internal/c/parts/audio/audio.cpp
a740g: Good idea. Done.

PR #569, opened 2024-11-19
Cleanup & refactor libqb

a740g:
This one is simple.

- Runs libqb and friends through clang-format.
- Removes many redundant forward declarations in libqb.cpp and qbx.cpp.
- Supresses some C++ compiler warnings.
- Stops `free()`ing `L->structure_base` in `qblist.cpp`.


internal/c/libqb/src/glut-message.h
```diff
@@ -72,58 +70,54 @@ class glut_message_get : public glut_message {
...
 
 class glut_message_set_window_title : public glut_message {
   public:
     char *newTitle;
     void execute();
 
-    glut_message_set_window_title(const char *title) : glut_message(false) {
-        newTitle = strdup(title);
-    }
+    glut_message_set_window_title(const char *title) : glut_message(false) { newTitle = strdup(title); }
```
mkilgore, line 110: Maybe we don't have to put `AllowShortFunctionsOnASingleLine` on :D IMO some of these one-line functions are too long for a single line.

mkilgore (approved):

internal/c/libqb/src/qbs__tostr.cpp
```diff
@@ -75,7 +95,10 @@ qbs *qbs__tostr(uint16_t value, int32_t digits, int32_t passed) {
     }
     return tqbs;
 }
+
 qbs *qbs__tostr(uint8_t value, int32_t digits, int32_t passed) {
+    (void)digits;
```
RhoSigma-QB64, line 100: What purpose serves these (void) casts on the passed function arguments?

internal/c/libqb/src/qbs__tostr.cpp
```diff
@@ -84,15 +107,18 @@ qbs *qbs__tostr(uint8_t value, int32_t digits, int32_t passed) {
...
 // floating points
 qbs *qbs__tostr(float value, int32_t digits, int32_t passed) {
     if (passed) {
         if (digits < 0) {
             error(QB_ERROR_ILLEGAL_FUNCTION_CALL);
             return qbs_new_txt("");
         }
-        if (digits < 1) digits = 1;
-        if (digits > 7) digits = 7;
+        if (digits < 1)
```
RhoSigma-QB64, line 118: IMO this is error prone for future extensions, when splitting single line if()s, then braces should be added for both, readability and safeguard in case more instructions are added into the condition later.

RhoSigma-QB64 (approved):

internal/c/libqb/src/glut-message.h
a740g: Thank you. I always found that weird to be honest. Changed it to `Empty` so that only empty functions get to be on the same line. Also added `SeparateDefinitionBlocks: Always` so that we always add an empty line after classes, structs, enums, and functions.

internal/c/libqb/src/qbs__tostr.cpp
a740g: I agree with you 100%. I used to omit braces for single line `if`s, but I’ve since changed that habit for the same reasons you mentioned. While there is a `clang-format` style that can add braces automatically, it’s not entirely safe to rely on it. Going forward, the best approach is to manually add braces when editing or reviewing code. For now, `clang-format` will simply handle indentation for single line `if`s.

PR #574, opened 2024-11-24
Implement _IIF and fix _ROR & _ROL

### LINKED ISSUE #403: Ternary Operator '?'
In a lot languages, the ones I can think of are C++, and C# a '?' question mark can be used to do ternary operators.
C++/C#: `result = (a >= 45) ? "true": "false";`
QB64?: `Print (a >= 45) ? "true": "false"`
It could make simple if statements easier, and stuff.
The options could be uses of function too like: `clamp(1, a, 10)` or something.
Just a more compact, and easier way to do smaller conditional things.

a740g:
This PR introduces `_IIF` support for QB64-PE, alongside bug fixes and improvements for `_ROL` and `_ROR`.

`_IIF` Implementation:
The `_IIF` pseudo-function allows for conditional evaluation like the ternary operator in C, with short-circuiting behavior:

```vb
result = _IIF(expression, truePart, falsePart)
```

Key features:

- Only the `truePart` or `falsePart` is evaluated, depending on the `expression`, mimicking C's conditional operator.
- The return type is determined by the type of `truePart`.
- Both parts cannot mix `STRING` with other types.

`_ROL` & `_ROR` fixes:

1. Resolved an issue where the second argument could incorrectly be a `STRING`.
2. Simplified the C-side logic using template functions for cleaner code.

Closes #403.


source/qb64pe.bas
```diff
@@ -23690,15 +23734,9 @@ FUNCTION SCase2$ (t$)
             CASE "_ANDALSO"
                 SCase2$ = "_AndAlso"
 
-            CASE "ANDALSO"
-                SCase2$ = "AndAlso"
-
```
RhoSigma-QB64, line 23695: Ah, you found some more $NOPREFIX remains, thought I catched them all in the meantime. :)

RhoSigma-QB64 (approved):

Issue #123, opened 2022-06-26
"Size" elements in the _MEM type should be _Unsigned Long or _Unsigned _Integer64 (depending on the arch)


a740g:
Currently the `_MEM` type is defined as follows. This makes it difficult to use `SIZE` and `ELEMENTSIZE` in code.

```VB
TYPE memory_type
    OFFSET AS _OFFSET       'start location of block(changes with byte position)
    SIZE AS _OFFSET         'size of block remaining at offset(changes with position)
    TYPE AS LONG            'type description of variable used(never changes)
    ELEMENTSIZE AS _OFFSET  'byte size of values inside the block(never changes)
    IMAGE AS LONG           'the image handle used when _MEMIMAGE(handle) is used
END TYPE
```

For example, the following code gives an error:

```VB
Dim p As _Mem
```
`...`
```VB
Dim s As _Integer64
s = p.SIZE ' SIZE is an _OFFSET and QB64 does not like this!
```

For 32-bit versions of QB64 `_MEM` should be defined as:
```VB
TYPE memory_type
    OFFSET AS _OFFSET
    SIZE AS _UNSIGNED LONG
    TYPE AS LONG
    ELEMENTSIZE AS _UNSIGNED LONG
    IMAGE AS LONG
END TYPE
```

For 64-bit versions of QB64 `_MEM` should be defined as:
```VB
TYPE memory_type
    OFFSET AS _OFFSET
    SIZE AS _UNSIGNED _INTEGER64
    TYPE AS LONG
    ELEMENTSIZE AS _UNSIGNED _INTEGER64
    IMAGE AS LONG
END TYPE
```

RhoSigma-QB64:
Since QB64 v2.0, hence also in ALL QB64-PE versions, `_CV` and `_MK$` can be used to convert `_OFFSET` values.

        a& = 0
        DIM m AS _MEM
        m = _MEM(a&)
        b& = _CV(LONG, _MK$(LONG, m.SIZE))
        PRINT b&

As far as I know it was discussed several times to change the entries, but for some good reasons it was not done. Instead the alterative with `_CV` and `_MK$` was implemented. I'll need to dig up the discussions in the old forum.

a740g:
Something similar was suggested by offbyone on a [Discord thread](https://discord.com/channels/592061709485735966/973039653076746322/990453493259182120). Then he suggested that I should submit an issue.

[...]

PR #580, opened 2024-11-27
_CLngPtr pseudo-function

a740g:
In QB64, the usage of `_OFFSET` variables in expressions is intentionally limited, with guardrails in place to discourage improper use. These restrictions are designed to prevent users from inadvertently performing invalid pointer arithmetic that could lead to reading or writing in unauthorized areas of memory. This is all good. However, there are scenarios where using _OFFSET variables in pointer arithmetic becomes necessary. For a detailed discussion, refer to issue #123.

Currently, these guardrails force users to rely on inefficient conversions or questionable hacks, both of which are undesirable.

Expensive conversion:
```vb

DIM m as _MEM

b&& = _CV(_INTEGER64, _MK$(_INTEGER64, m.SIZE))
```

Hack:

```vb
$IF 32BIT THEN
    FUNCTION CLngPtr~& ALIAS "uintptr_t" (BYVAL o AS _OFFSET)
$ELSE
    FUNCTION CLngPtr~&& ALIAS "uintptr_t" (BYVAL o AS _OFFSET)
$END IF 
```

This PR introduces the `_CLngPtr` pseudo-function, which allows users to directly retrieve the value of an _OFFSET variable and allows its usage in expressions where usage of _OFFSET was previously not allowed. Internally, no actual function call is performed. The name is inspired by Microsoft's [CLngPtr](https://learn.microsoft.com/en-us/office/vba/language/concepts/getting-started/type-conversion-functions), which does something similar.

```vb
result = _CLngPtr(offset%&)
```


RhoSigma-QB64 (approved):
Just don't know if I like the name. When reading `_CLngPtr` my immediate thought is more in the direction of obtaining a pointer from the given `LONG` variable, hence `_OFFSET(lvar&)` or in C `ptrszint = &somelongvar`. The name is also too similar to `CLNG` which does something completely different.

But as this pseudo function effectively just makes an explicit type cast I'd probably prefer something more straight forward to that behavior, e.g. simply `_OffToInt`, `_OffsetAsNumber`, `_OffsetVal` or something similar.

a740g:
I have had trouble picking the name, eventually settling for something Microsoft uses, thinking it may be familiar. But I think I like @mkilgore's suggestion better now. A `_Cast ` command would be more useful.

Honestly, I am unsure why we have restrictions on _OFFSET. However, after looking at the code, I found multiple places where it is checked if we are dealing with an _OFFSET or not. Notably:
 https://github.com/QB64-Phoenix-Edition/QB64pe/blob/bb829766fa9cfe1bdc0427cda937e9f88947de84/source/qb64pe.bas#L17965
https://github.com/QB64-Phoenix-Edition/QB64pe/blob/bb829766fa9cfe1bdc0427cda937e9f88947de84/source/qb64pe.bas#L10772
https://github.com/QB64-Phoenix-Edition/QB64pe/blob/bb829766fa9cfe1bdc0427cda937e9f88947de84/source/qb64pe.bas#L16809
https://github.com/QB64-Phoenix-Edition/QB64pe/blob/bb829766fa9cfe1bdc0427cda937e9f88947de84/source/qb64pe.bas#L17472
And there are a few more. I did consider removing those. But I am not really sure what that'll break.

mkilgore:
I think in general the `_OFFSET` restrictions make sense _if_ we consider it to just be a pointer type, like a `char *`. Storing regular integers in a `char *` is a messy way to do things and makes it easy to get values mixed up and deref a non-pointer type. Keeping a clean separation helps avoid big errors, and in C and C++ storing numbers in a `char *` is almost never necessary so it's not a big deal.

The issue for QB64 is that it breaks this thinking with `_MEM.SIZE`, because that value is obviously not a pointer but still has the `_OFFSET` type. If we could change `_MEM.SIZE` (and the other entries) to a new variable type in a non-breaking way then that would be the best solution, but I'm pretty sure we can't do that without breaking existing code (it might be worth a try though). If we can't introduce a new type, then removing the restrictions of `_OFFSET` and allowing it to act like both a `char *` and `size_t` probably makes the most sense.

a740g:
Closing this based on discussion above. Will open a fresh PR with _OFFSET restrictions removed and add new tests.

PR #602, opened 2025-01-24
Disables mouse event suppression on macOS when the mouse is hidden and moved

a740g:
This PR fixes a bug on macOS where mouse events are suppressed for 0.25 seconds when `_MOUSEMOVE` is used. The issue is resolved by calling `CGAssociateMouseAndMouseCursorPosition()`. Note that the event suppression bypass only happens when the mouse is hidden.

Additionally, as discussed on Discord, this PR prevents `_FULLPATH$` from throwing an error when an empty string is passed.

### DIFF
--- a/internal/c/libqb.cpp
+++ b/internal/c/libqb.cpp
@@ -28826,9 +28826,8 @@ int main(int argc, char *argv[]) {
     singlespace = qbs_new_cmem(1, 0);
     singlespace->chr[0] = 32;
 
-    // store _CWD$ for recall using _STARTDIR$ in g_startDir
-    g_startDir = qbs_new(0, 0);
-    qbs_set(g_startDir, func__cwd());
+    // Store _CWD$ for recall using _STARTDIR$
+    FS_SaveStartDirectory();
 
 // switch to directory of this EXE file
 // http://stackoverflow.com/questions/1023306/finding-current-executables-path-without-proc-self-exe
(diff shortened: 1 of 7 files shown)


RhoSigma-QB64 (approved):
Thanks for the `_FULLPATH$` change, maybe you wanna change this too https://discord.com/channels/975381912350752819/975389657212805160/1329530280155021443 as according to the discussion all supported platforms don't care about double separators, this wouldn't be a breaking change.

grymmjack:
✅ Tested and working great on this mac:
image

I ran the example you shared in the discord thread ahead of checking out and trying with the PR code and this new code does seem more responsive/fluid.

Both had trackpad intertia - I didn't try with mouse just trackpad.

a740g:
> Thanks for the `_FULLPATH$` change, maybe you wanna change this too https://discord.com/channels/975381912350752819/975389657212805160/1329530280155021443 as according to the discussion all supported platforms don't care about double separators, this wouldn't be a breaking change.

No problem. Done. I ran some tests and it looks like we are good on all platforms.

a740g:
> ✅ Tested and working great on this mac: image
> 
> I ran the example you shared in the discord thread ahead of checking out and trying with the PR code and this new code does seem more responsive/fluid.
> 
> Both had trackpad intertia - I didn't try with mouse just trackpad.

Thank you @grymmjack

PR #606, opened 2025-02-01
Refactor HSB color functions

RhoSigma-QB64:
- removed the HSB functions from BASIC level in afterlastline.bm
- translated the functions to C/C++ level and put into libqb.cpp


a740g (approved):
You may want to move these to `internal/c/libqb/include/image.h` and `internal/c/parts/video/image/image.cpp`. Both `libqb.cpp` and `qbx.cpp` pull in `image.h`.

PR #612, opened 2025-02-20
Meta format off on

RhoSigma-QB64:
Further improve auto formatting behavior, so we can have proper formatting switched on by default, but can suppress any auto spacing (indention+single spacing of code elements) for sections we rather want to space/align manually for better readability.

Implemented as legacy meta-commands after a REM or apostrophe:
- `'$FORMAT:OFF` will suppress indention and single-spacing code elements until switched on again
- `'$FORMAT:ON` to restore back to default formatting as set in the IDE Code Layout dialog or given as -f commandline switches in conjunction with the -y formatting mode


source/qb64pe.bas
```diff
@@ -20792,7 +20794,14 @@ FUNCTION lineformat$ (a$)
         '    : metacommands do not need to be terminated by word boundaries.
         '      E.g., $STATICanychars$DYNAMIC is valid
 
-        IF MID$(c$, x, 7) = "$STATIC" THEN
+        IF MID$(c$, x, 11) = "$FORMAT:OFF" THEN
+            MID$(layoutcomment$, x + cdif, 11) = SCase$("$Format:Off")
+            IDEAutoIndent = 0: IDEAutoLayout = 0
```
mkilgore, line 20799: What happens if there's an unpaired `'$Format:Off`, do the `DEF` values get restored somewhere after the compilation is done?

RhoSigma-QB64: Good point, and that's not the only problem, I also need to do further adjustments to avoid changed formatting states in $INCLUDE files leaking back into the main code and proper highligting for meta after comment is also somewhat more complicated than just adding the keyword to the list, it also needs some special handling in the syntax highlighter itself.

SteveMcNeill: IF MID$(c$, x, 11) = "$FORMAT:OFF" ...
 
 It would be better to break this down into a single statement and then parse it for incorrect spacing
 
IF MID$(c$, x, 8) = "$FORMAT:"  THEN
     r$ = _TRIM$(MID$(c$,9))
     IF r$ = "ON" THEN
     ...
    ELSEIF r$ = "OFF THEN
    ...
    ELSE
        e$ = "Error Error Will Robinson!!": GOTO errmes
    END IF
END IF
 
Thus an user could type:
$FORMAT: ON
$FORMAT:ON
$FORMAT:    ON

And all would read without an error and then you can set them to the proper layout that you want for that command.

RhoSigma-QB64: We don't even allow that for the regular metacommands. $CHECKING:ON, $RESIZE:ON, $COLOR:32, $CONSOLE:ONLY, $ASSERTS:CONSOLE etc, all are required to be written in one solid block, so why we should start that now for a pseudo metacommand?

SteveMcNeill: Honestly, I don't see why all the commands aren't set up in such a manner.  The change would be minimal in the source, but it'd just make things more flexible and convenient for the end user.  Why toss an error over an extra space, when it's obvious what the intent is?  IMHO, it's always a good thing to make the language/IDE more intelligent and flexible when we can.

For example:```
If a3u$ = "$RESIZE:OFF" Then
    layout$ = SCase$("$Resize:Off")
    ResizeOn = 0: ResizeScale = 0
    GoTo finishednonexec
End If
If a3u$ = "$RESIZE:ON" Then
    layout$ = SCase$("$Resize:On")
    ResizeOn = 1: ResizeScale = 0
    GoTo finishednonexec
End If
If a3u$ = "$RESIZE:STRETCH" Then
    layout$ = SCase$("$Resize:Stretch")
    ResizeOn = 1: ResizeScale = 1
    GoTo finishednonexec
End If
If a3u$ = "$RESIZE:SMOOTH" Then
    layout$ = SCase$("$Resize:Smooth")
    ResizeOn = 1: ResizeScale = 2
    GoTo finishednonexec
End If
```

This would easily convert over to: ```
If Left$(a3u$, 8) = "$RESIZE:" Then
    Select Case _Trim$(Mid$(a3u$, 9))
        Case "OFF"
            layout$ = SCase$("$Resize:Off")
            ResizeOn = 0: ResizeScale = 0
            GoTo finishednonexec
        Case "ON"
            layout$ = SCase$("$Resize:On")
            ResizeOn = 1: ResizeScale = 0
            GoTo finishednonexec
        Case "STRETCH"
            layout$ = SCase$("$Resize:Stretch")
            ResizeOn = 1: ResizeScale = 1
            GoTo finishednonexec
        Case "SMOOTH"
            layout$ = SCase$("$Resize:Smooth")
            ResizeOn = 1: ResizeScale = 2
            GoTo finishednonexec
    End Select
End If
```

It wouldn't break anything old, would only make it easier in the future, and is still just as simple to read/understand the workflow here, as it was before.  I really can't see any negative to parsing the commands in such a manner, and see just added flexibility here.  ;)

SteveMcNeill: (Note also the improved efficiency in the above.  The first has to check a3u$ 4 different times to see if a $RESIZE metacommand is in effect.  The 2nd only checks once for $RESIZE of any sort and then moves on, only checking after that to see what type of resize is in effect.  Since 99.9% of code isn't going to be this one command, that's a lot of IF checks that we're skipping when parsing that code repeatedly.  

Anything that might help improve IDE lag/efficiency is also a good thing, IMHO.)

a740g (approved):

PR #614, opened 2025-03-04
Audio update and cleanup

a740g:
This PR includes the following updates and improvements:

- Updates `miniaudio`, `TinySoundFont`, `QOA`, `Libxmp-lite`, `foo_midi`, and `libmidi`.
- Fixes playback-related issues in `TSFPlayer.cpp` and `mod_ma_vtable.cpp`.
- Converts `InstrumentBankManager` into a singleton to eliminate and hide global variables.
- Converts `AudioEngine` into a singleton, removing the need for `snd_init()` and `snd_un_init()`.
- Adopts the new `miniaudio` dual-file (`.c` + `.h`) setup.
- Organizes libraries into separate directories.
- Adds sample rate support to `_SndNew()`.
- Moves all information, warning, and trace logs to use the new `libqb-logger`.


RhoSigma-QB64 (approved):
Additionally, may I ask you for a favor, could you please have a look here https://github.com/QB64-Phoenix-Edition/QB64pe/blob/main/source/qb64pe.bas#L9400

The `typ$` at the very end of the line needs to be enclosed in an `SCase2$()` call. The ommision prevents the correct keyword formatting of variable types given in the `_MEMPUT block, offset, value AS TYPE` syntax, but it's not worth to make an own PR for it.

a740g:
> Additionally, may I ask you for a favor, could you please have a look here https://github.com/QB64-Phoenix-Edition/QB64pe/blob/main/source/qb64pe.bas#L9400
> 
> The `typ$` at the very end of the line needs to be enclosed in an `SCase2$()` call. The ommision prevents the correct keyword formatting of variable types given in the `_MEMPUT block, offset, value AS TYPE` syntax, but it's not worth to make an own PR for it.

I think this one as well, right?

https://github.com/QB64-Phoenix-Edition/QB64pe/blob/7c59a33cfd05db39f3e965a65d1eadcafd91c3a6/source/qb64pe.bas#L9507

PR #615, opened 2025-03-05
Fix log levels

a740g:
Nothing serious. I messed up the log levels in the last PR mostly using info logs instead of trace. This resulted in polluting the log output. This PR moves most logs to trace and warn levels using info sparingly.

### DIFF
--- a/internal/c/parts/audio/audio.cpp
+++ b/internal/c/parts/audio/audio.cpp
@@ -3714,7 +3714,7 @@ class AudioEngine {
                 break;
 
             default:
-                audio_log_trace("Unknown format");
+                audio_log_warn("Unknown format");
                 return;
             }
         } else {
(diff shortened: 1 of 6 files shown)


RhoSigma-QB64 (approved):
Did the latest changes introduce a new scope into the logging system such as `audio` maybe? Then that should be noted here: https://qb64phoenix.com/qb64wiki/index.php/Logging#Scopes

a740g:
> Did the latest changes introduce a new scope into the logging system such as `audio` maybe? Then that should be noted here: https://qb64phoenix.com/qb64wiki/index.php/Logging#Scopes

I agree. We need to update the wiki. However, the audio scope has been available since v4.0.0. It seems we forgot to mention it in the wiki.

mkilgore:
> > Did the latest changes introduce a new scope into the logging system such as `audio` maybe? Then that should be noted here: https://qb64phoenix.com/qb64wiki/index.php/Logging#Scopes
> 
> I agree. We need to update the wiki. However, the audio scope has been available since v4.0.0. It seems we forgot to mention it in the wiki.

I'm pretty sure I didn't mention it in the Wiki because it wasn't in use in v4.0.0.

Issue #608, opened 2025-02-07
Replace GLUT with something like GLFW / RGFW


a740g:
We've had this conversation multiple times on Discord, and we all agreed to some extent that we need to replace [Freeglut](https://github.com/freeglut/freeglut) with something better and more modern like [GLFW](https://github.com/glfw/glfw). I thought it would be best if we put this formally in the issue tracker and continue our discussion here. That way, we can better track which way we want to go ahead with this.

As it stands, multiple things need to be done to successfully remove [Freeglut](https://github.com/freeglut/freeglut) from our codebase. Some of these things are intricate and we also need to ensure we are not breaking anything along the way (especially stuff to do with the keyboard and mouse).

That said, removing [Freeglut](https://github.com/freeglut/freeglut) has multiple benefits. Here are a few things that I can think of after having experimented with [RGFW](https://github.com/ColleagueRiley/RGFW) with a fair amount of success.
 
- The main loop will be in our control, and we will not need to resort to ugly hacks (like modifying the library and plugging in our code).
- We will not be at the mercy of the [Freeglut](https://github.com/freeglut/freeglut) implementation of macOS.
- A lot of "Windows-only" stuff will go away. For example, we will get drag-and-drop support for all OSes.
- We can remove some third-party libraries and we'll be able to simplify a lot of things.  For example, [GLFW](https://github.com/glfw/glfw)/[RGFW](https://github.com/ColleagueRiley/RGFW) has a built-in game controller, drag-and-drop, and clipboard support.
- This should also allow us to take care of #292, #283, #128, and #271.

PR #616, opened 2025-03-09
Fix _MOUSEX and _MOUSEY when the mouse pointer is hidden on macOS

a740g:
This partially reverses the approach taken in #602. Using `CGAssociateMouseAndMouseCursorPosition()` was a mistake - it fixed `_MOUSEMOVEMENTX` and `_MOUSEMOVEMENTY` for FPS-style camera movement but broke `_MOUSEX` and `_MOUSEY` by preventing GLUT from triggering the `glutMotionFunc` callback while the mouse pointer was hidden.

To address this, we now use `CGEventSourceSetLocalEventsSuppressionInterval()` to work around macOS’s 0.25-second event suppression. However, this means `_MOUSEMOVEMENTX` and `_MOUSEMOVEMENTY` remain broken because macOS generates artificial movement events when `glutWarpPointer` is used.

For now, this is an acceptable trade-off. A proper long-term fix depends on implementing #608.

Many thanks to `NakedApe` from the QB64-PE forum for pointing out the issue. See [this](https://qb64phoenix.com/forum/showthread.php?tid=3290).

### DIFF
--- a/internal/c/libqb.cpp
+++ b/internal/c/libqb.cpp
@@ -28251,7 +28251,7 @@ void GLUT_MOTION_FUNC(int x, int y) {
     int32 xrel, yrel;
 
     // This is used to save the last mouse position which is then paired with the mouse wheel event on macOS
-    macMouseUpdatePosition(x, y);
+    MacMouse_UpdatePosition(x, y);
 
     mouse_message_queue_struct *queue = &mouse_message_queue;
 
--- a/internal/c/libqb/include/mac-mouse-support.h
+++ b/internal/c/libqb/include/mac-mouse-support.h
@@ -1,21 +1,10 @@
 #pragma once
 
 #ifdef QB64_MACOSX
-void macMouseInit();
-void macMouseDone();
-void macMouseUpdatePosition(int x, int y);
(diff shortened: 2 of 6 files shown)


RhoSigma-QB64 (approved):
And I'm afraid you also have to revisit the _MEMPUT/_MEMFILL formatting fix here https://github.com/QB64-Phoenix-Edition/QB64pe/pull/614/commits/f3cf4c2a9b0f2cf2e35572b17d16945de24ff4c8

It must be an `SCase2$()` call, not a `SCase$()` call.

a740g:
> And I'm afraid you also have to revisit the _MEMPUT/_MEMFILL formatting fix here [f3cf4c2](https://github.com/QB64-Phoenix-Edition/QB64pe/commit/f3cf4c2a9b0f2cf2e35572b17d16945de24ff4c8)
> 
> It must be an `SCase2$()` call, not a `SCase$()` call.

🤦‍♂️I swear I was thinking about `SCase2$()` while making those changes. 😄 I'll update the `SCase$()` to `SCase2$()` and push the changes.

PR #626, opened 2025-04-14
Support Using System MinGW on Windows

a740g:
This change adds support for using a system-installed compiler on Windows, such as a pre-existing MinGW toolchain (from MSYS2, Winget etc.). This can be helpful in situations where the user either cannot or prefers not to use the compiler bundled with QB64-PE.

A clear warning is displayed when this option is enabled to ensure users understand the implications. This feature is available exclusively on Windows.

### DIFF
--- a/source/qb64pe.bas
+++ b/source/qb64pe.bas
@@ -89,6 +89,7 @@ IF _DIREXISTS("internal") = 0 THEN
 END IF
 
 DIM SHARED Include_GDB_Debugging_Info 'set using "config.ini" or "Compiler settings" dialog
+DIM SHARED UseSystemMinGW AS _BYTE
 
 DIM SHARED DEPENDENCY_LAST
 CONST DEPENDENCY_LOADFONT = 1: DEPENDENCY_LAST = DEPENDENCY_LAST + 1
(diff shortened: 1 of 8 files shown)


source/qb64pe.bas
```diff
@@ -13414,6 +13419,7 @@ SUB CMDLineTemporarySettingsHelp
     PRINT "  -f:ExtraLinkerFlags=[string]          Extra flags for the Linker"
     PRINT "  -f:MaxCompilerProcesses=[integer]     Max C++ Compiler processes to use"
     PRINT "  -f:GenerateLicenseFile=[true|false]   Produce a license.txt file for program"
+    PRINT "  -f:UseSystemCompiler=[true|false]     Use the system C++ compiler instead of the bundled one (Windows only)"
```
RhoSigma-QB64, line 13422: Description to long, please keep in 80 columns bound, i.e. break the line after the "of" abd continue indented beneth the "Use". See options -e or -l for example. Everything else looks good to me.

RhoSigma-QB64 (approved):

source/qb64pe.bas
a740g: Done.

PR #627, opened 2025-04-23
Fix IDE error when loading large source file

FellippeHeitor:
The IDE would crash when a large source file was loaded and the syntax highlighter had to build the list of sub/functions on its own, due to errors in the loaded file preventing compilation from going through.

### DIFF
--- a/source/ide/ide_methods.bas
+++ b/source/ide/ide_methods.bas
@@ -12689,6 +12689,10 @@ SUB ideshowtext
                 manualList = -1
                 listOfCustomKeywords$ = LEFT$(listOfCustomKeywords$, customKeywordsLength)
                 FOR y = 1 TO iden
+                    DO UNTIL y < UBOUND(InvalidLine)
+                        REDIM _PRESERVE InvalidLine(UBOUND(InvalidLine) + 1000) AS _BYTE
+                    LOOP
+
                     IF InvalidLine(y) <> 0 THEN _CONTINUE
                     a$ = UCASE$(_TRIM$(idegetline(y)))
                     sf = 0


FellippeHeitor:
BTW, to reproduce the issue before this patch is applied, just load ide_methods.bas into the ide itself. v4.1.0.

SteveMcNeill:
why not make this a single line of logic?

```
                    DO UNTIL y < UBOUND(InvalidLine)
                        REDIM _PRESERVE InvalidLine(UBOUND(InvalidLine) + 1000) AS _BYTE
                    LOOP
```

Instead: ```
    REDIM _PRESERVE InvalidLine(y + 1000) AS _BYTE
```

No need for a DO... LOOP here, is there?

**Comment by FellippeHeitor** [comment]
Removing the DO/LOOP block, which serves as a condition check here, would increase the array indefinitely, eventually leading to a memory issue. I just copied the block of code you wrote back when you implemented InvalidLine() logic back in the day (same block is present in ide_methods, ~ line 12800-ish:

```
                DO UNTIL l < UBOUND(InvalidLine) 'make certain we have enough InValidLine elements to cover us in case someone scrolls QB64
                    REDIM _PRESERVE InvalidLine(UBOUND(InvalidLine) + 1000) AS _BYTE '   to the end of a program before the IDE has finished
                LOOP '                                                      verifying the code and growing the array during the IDE passes.
```

**Comment by SteveMcNeill** [comment]
```
IF I > UBOUND(InvalidLine) THEN  REDIM _PRESERVE InvalidLine(y + 1000) AS _BYTE
```

No need for the DO.. Loop as it's just repeating that one process over and over until it grows bigger than y by 1000.  The other segment can be shortened and cleaned up as well.   

We all learn better tricks as we keep learning and growing as programmers.  ;)

FellippeHeitor:
Your logic for using the DO/LOOP then is still valid, I wouldn't remove it from the other block, especially because it's used for cases when the user may have scrolled way beyond the current ubound() of the array, making a 1000 increase insufficient. Trust old Steve on that one. Anyway, even if the patch doesn't make it as is, please look into the issue. It's broken for now.

a740g (approved):

PR #639, opened 2025-06-05
Use _TRUE/_FALSE constants where applicable...

RhoSigma-QB64:
... also slightly renamed some variables (make it camel case) and removed some unused (commented) code

### DIFF
--- a/internal/support/vwatch/vwatch.bm
+++ b/internal/support/vwatch/vwatch.bm
@@ -46,7 +46,7 @@ SUB vwatch (globalVariables AS _OFFSET, localVariables AS _OFFSET)
         vw_cmd$ = "me:" + COMMAND$(0)
         GOSUB SendCommand
         IF _WINDOWHANDLE THEN
-            vw_hwndSent = -1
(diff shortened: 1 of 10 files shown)


internal/support/vwatch/vwatch.bm
```diff
@@ -151,15 +151,15 @@ SUB vwatch (globalVariables AS _OFFSET, localVariables AS _OFFSET)
         EXIT SUB
     END IF
 
-    IF vwatch_linenumber = vw_lastLine AND vw_setNextLine = 0 THEN EXIT SUB
```
FellippeHeitor, line 154: Rather than a _FALSE value (hence the use of NOT, I assume), vw_setNextLine will range from 0 to the total lines of the user's program. NOT (some other value) wouldn't be the same as NOT -1 (= 0). This looks like it could cause issues.

FellippeHeitor: I stand corrected. It's `vwatch_goto` that needs to go higher. Been a while.

a740g (approved):

PR #641, opened 2025-06-10
Replace wmic with powershell equivalent in Windows set-up scripts.

SteveMcNeill:
These should be more forward and backward compatible for us.  (I hope.)  Powershell here is basically just doing the same call to the Win32_Processor and getting the exact same information for us as before.  We're just accessing it with another command, rather than the old wmic command which Windows is depreciating and maybe obsoleting soon.


spriggsyspriggs:
The PowerShell calls should have "NoProfile" switch added so as to speed up each call. Otherwise, PowerShell will be loading the user profile each time.

PR #676, opened 2026-01-30
Color enhancement

RhoSigma-QB64:
- we may `_(U)PRINTSTRING` to an optional destination image, but could not set the printing colors the same easy way until now
- new syntax `COLOR [foreground&][, background&][, , destHandle&]`


flukiluke (commented):
It would be good to have a few basic tests to confirm the syntax and _dest selection logic works as expected.

PR #696, opened 2026-03-13
Support for static nested arrays in TYPE (including _MEM, UBOUND, LBOUND, ERASE)

QB64Petr:
Description
Static arrays within arrays in TYPE are now fully supported. In previous versions, only declaration was possible, but many core functions were missing. 

 Key Improvements:
Supports 1D and Multi-D static nested arrays.
Added support for: `_Mem`, `_MemGet`, `_MemPut`, `_MemFill`, `Lbound`, `Ubound`, and `Erase`.
`Erase` now allows clearing either the entire parent array or specifically just the nested array.

Installation / Testing
The PR includes updated core files and a set of test BAS programs in the `/programs/tests/` folder.


mkilgore (changes requested):
It looks like you changed the line endings in qb64pe.bas, please undo that commit and add the file back with the correct line endings. It treats that like you changed every line so the diff is impossible to review and the history gets hard to follow.

QB64Petr:
I wrote upgrade to source code from https://github.com/QB64-Phoenix-Edition/QB64pe/blob/main/source/qb64pe.bas  

I still need to figure out how to reliably get the latest version. I also need to add support for ReDim and Option Base – I completely forgot about those. I'm on it.

Right now, the priority is figuring out how to make sure I'm working with the latest code. I really thought I had the right version.

QB64Petr:
II already figured out how to get the fresh version. I'm still learning with GitHub. Let me know if adding the missing updates would be too complicated and I'll just rewrite it. It's my fault again.

QB64Petr:
@mkilgore I start from this version:  https://github.com/QB64-Phoenix-Edition/QB64pe/blob/main/source/qb64pe.bas

a740g:
Replaced by #699.

PR #697, opened 2026-03-15
Add support for static nested UDT member arrays and compile tests

QB64Petr:
This patch adds support for static nested member arrays in QB64PE and
includes matching tests under tests/compile_tests.

The work covers:
- parsing and handling of static member arrays inside TYPE blocks
- REDIM and REDIM _PRESERVE support for nested static member arrays
- OPTION BASE handling for member arrays declared without an explicit
  lower bound
- support for array declarators in "AS type element-list" syntax inside
  TYPE blocks
- related member-array handling in ERASE, LBOUND/UBOUND, _OFFSET and
  _MEM-family operations

ERASE semantics for static nested member arrays are clarified as well.
ERASE clears the contents of the member array, but does not remove or
resize it. Its bounds remain unchanged because the array is part of the
containing UDT layout.

One previous test was corrected because its expectation was wrong.
After ERASE, a fixed-length string member does not become an empty
string; it keeps its fixed length and its contents become CHR$(0) bytes.
Erase allow delete the main field. This will also delete the nested fields.


SteveMcNeill:
I haven't had a chance to dig into this yet; I've been working on the forums and spam bots and crawlers, and then getting ready for life here on the farm as the colder weather slowly fades away.   One thing I wanted to stop in and make certain to do, however, was offer you an invite to Discord with us Petr.  You'll find at least some of us on there most of the time, and it's a very good way for instant feedback and tossing of ideas back and forth with stuff.  You can also see what other folks are up to with development, where you might help, where they might help you, and generally just chat and goof off with us when we're not doing anything productive.  (Which is most of the time for me anymore, to be honest.)

https://discord.gg/D2M7hepTSx  <-- You is invited to Discord with us!  It'll be a pleasure to have you around and to get to chat/work with you more often there.

a740g:
Replaced by #699.

PR #699, opened 2026-03-16
Nested Static Arrays in TYPE

QB64Petr:
Support for nested static fields in type - fixes. More fixes to follow!


tests/compile_tests/arrays/08_redim_same_bounds_option_base1_classic_udt.bas
```diff
@@ -0,0 +1,29 @@
+$CONSOLE:ONLY
+Option _Explicit
+Option Base 1
+
+Type T
+    Values( 3) As Long
+End Type
+
+ReDim reference(3) As Long
+ReDim x(0) As T
```
mkilgore: I think this hits an error with `OPTION BASE 1`.

QB64Petr:
why https://github.com/QB64-Phoenix-Edition/QB64pe/actions/runs/23353355644/job/67938536092?pr=699 run not next test?

source/utilities/type.bi
```diff
@@ -71,5 +71,10 @@ REDIM SHARED udtesize(1000) AS LONG
 REDIM SHARED udtetype(1000) AS LONG
 REDIM SHARED udtetypesize(1000) AS LONG
 REDIM SHARED udtearrayelements(1000) AS LONG
+REDIM SHARED udtearraybase(1000) AS LONG
+REDIM SHARED udtearraydims(1000) AS LONG
+' Serialized per-dimension bounds for static array members stored inline inside a TYPE.
```
mkilgore, line 76: We should document the format of this string. We could also probably move the functions for working with the descriptor string into a separate .bas file.

source/qb64pe.bas
```diff
@@ -1,4 +1,4 @@
-'All variables will be of type LONG unless explicitly defined
+'All variables will be of type LONG unless explicitly defined         002
```
mkilgore: The 002 can be removed.

source/qb64pe.bas
```diff
@@ -1997,6 +2025,28 @@ DO
...
+                                        DO WHILE i3 <= n
+                                            l$ = getelement$(a$, i3)
+                                            IF l$ = "(" THEN b2 = b2 + 1
+                                            IF l$ = ")" THEN
+                                                b2 = b2 - 1
+                                                IF b2 = 0 THEN EXIT DO
+                                            END IF
+                                            i3 = i3 + 1
+                                        LOOP
+                                        IF b2 <> 0 THEN a$ = "Expected)": GOTO errmes
```
mkilgore: There needs to be a space between `Expected` and `)`. This goes for all the cases where `"Expected)"` shows up.

source/qb64pe.bas
```diff
@@ -14249,6 +14575,268 @@ FUNCTION arrayreference$ (indexes$, typ)
...
+    IF comma_pos = 0 THEN EXIT FUNCTION
+
+    lower_bound = VAL(LEFT$(pair$, comma_pos - 1))
+    element_count = VAL(MID$(pair$, comma_pos + 1))
+    IF element_count <= 0 THEN EXIT FUNCTION
+
+    ParseNextUDTArrayDescriptorDim = -1
+END FUNCTION
+
+'Preserves the original ParseUDTArrayBounds() behavior for existing callers that should
```
mkilgore: These two comments don't make sense since these functions didn't exist before this change.

Let's change the comments on both functions. This one should clarify it does the same thing as `ParseUDTArrayBoundsEx&()` but uses the global `optionbase` setting. The `ParseUDTArrayBoundsEx&()` comment should clarify what the function actually does.

source/qb64pe.bas
```diff
@@ -678,6 +682,12 @@ DIM SHARED glinkid AS LONG
 DIM SHARED glinkarg AS INTEGER
 DIM SHARED typname2typsize AS LONG
 DIM SHARED uniquenumbern AS LONG
+' When this flag is enabled, a TYPE member that is declared as a static array may be referenced
+' without an explicit element index. That special form is only valid in contexts that need the
+' whole inline array block, such as LBOUND/UBOUND, ERASE, _OFFSET and the _MEM family.
+' Normal expression evaluation keeps this disabled so that legacy "Expected array index" errors
```
mkilgore: nit: They're not "legacy", just in some contexts an index is required while in others it's not.

source/qb64pe.bas
```diff
@@ -1355,6 +1368,7 @@ addmetastatic = 0
 addmetadynamic = 0
 DynamicMode = 0
 optionbase = 0
+udtparse_optionbase = 0
```
mkilgore: Is there a reason to make this separate from the global `optionbase`? It seems to complicate things. The only catch is we have to reset it after the first parse step. If we reused it then we wouldn't need the separate `ParseUDTArrayBoundsEx` function.

source/qb64pe.bas
```diff
@@ -8100,13 +8226,63 @@ DO
...
+            targetref$ = evaluate(var$, targettyp)
+            udt_allow_bare_array = 0
+            IF Error_Happened = 0 THEN
+                IF (targettyp AND ISREFERENCE) <> 0 AND (targettyp AND ISARRAY) <> 0 THEN
+                    s1 = INSTR(targetref$, sp3)
+                    IF s1 THEN s2 = INSTR(s1 + LEN(sp3), targetref$, sp3) ELSE s2 = 0
+                    IF s2 THEN s3 = INSTR(s2 + LEN(sp3), targetref$, sp3) ELSE s3 = 0
+                    IF s3 THEN
+                        idnum = VAL(LEFT$(targetref$, s1 - 1))
+                        E = VAL(MID$(targetref$, s2 + LEN(sp3), s3 - (s2 + LEN(sp3))))
```
mkilgore: In general it would be good to avoid using the variable `E` because that's used by the error handler.

source/qb64pe.bas
```diff
@@ -8100,13 +8226,63 @@ DO
...
 
             erasedarray:
-            IF i < n THEN
-                i = i + 1: n$ = getelement$(a$, i): IF n$ <> "," THEN a$ = "Expected ,": GOTO errmes
+            IF i <= n THEN
+                n$ = getelement$(a$, i)
+                IF n$ <> "," THEN a$ = "Expected,": GOTO errmes
                 l$ = l$ + sp2 + ","
-                i = i + 1: IF i > n THEN a$ = "Expected , ...": GOTO errmes
+                i = i + 1: IF i > n THEN a$ = "Expected, ...": GOTO errmes
```
mkilgore: Space was removed

source/qb64pe.bas
```diff
@@ -15715,11 +16303,28 @@ FUNCTION udtreference$ (o$, a$, typ AS LONG)
     udtfindelenext:
     IF getelement$(a$, i) <> "." THEN Give_Error "Expected .": EXIT FUNCTION
     i = i + 1
-    n$ = getelement$(a$, i)
+    n$ = UCASE$(getelement$(a$, i))
```
mkilgore: Why does this need UCASE?

source/qb64pe.bas
```diff
@@ -15742,6 +16347,60 @@ FUNCTION udtreference$ (o$, a$, typ AS LONG)
...
+            IF incmem THEN typ = typ + ISINCONVENTIONALMEMORY
+            EXIT FUNCTION
+        END IF
+
+        arrindex$ = UDTArrayIndexExpr$(memberindexes$, udtearraydims(E), udtearraydesc(E))
+        IF Error_Happened THEN EXIT FUNCTION
+        IF o MOD 8 THEN Give_Error "Non-byte aligned user defined type": EXIT FUNCTION
+        arrbytes = (udtesize(E) \ 8) \ udtearrayelements(E)
+        o = o \ 8
+        IF o$ <> "" THEN
```
mkilgore: Rather than have all these special cases for `o$` and `o` being blank or zero, I'd rather we just set `o$` to `"0"` and allow for zero's in the output. Ultimately the compiler will remove the redundant zero's, I can't imagine we're saving all that much to justify the added repetition in these special cases.

QB64Petr:
Thanks for the review.

I’m currently in the middle of another chunk of work, so I’m going to pause this PR for a bit rather than context-switch back and forth. I’ve seen the requested small fixes and comment additions, and I’ll come back to this PR and address them in one pass once I finish the current work.

So this PR is paused for now, not abandoned.


04-04-2026    I continue to work on the repairs

source/qb64pe.bas
QB64Petr: This is separate on purpose. Normal arrays and nested arrays do not go through exactly the same handling, so sharing one Option Base variable would couple two different cases and make bugs more likely. Using a separate variable for nested arrays keeps that logic isolated and avoids interfering with normal array behavior.

source/qb64pe.bas
QB64Petr: changed, E is not used anymore

source/qb64pe.bas
QB64Petr: Yes, I filled in the spaces everywhere in new version, I'll send that version in a moment.

source/qb64pe.bas
QB64Petr: It's there because otherwise the parser has trouble recognizing the last element of the nested array.

source/qb64pe.bas
QB64Petr: I partly agree, and the new version already simplifies this by normalizing blank o$ to "0" in udtreference$.

That said, this is not just cosmetic output cleanup. In this area the normalized shape of o$ is also used by later compiler logic for nested-array recognition. That is why the same change also required updates in the whole-array detection and LEN handling.

So simplification is fine where it is safe, but o$ is not just dead text that only the C compiler sees later.

source/utilities/type.bi
QB64Petr: it is now in new source code

QB64Petr:
Please review this PR.
It has been sitting open for three weeks, and I have already rebased/updated it and resolved the conflict with upstream.

The implementation is tested and ready for review.
At this point, I need a concrete outcome: merge it, request specific changes, or explicitly reject it.

Please do not leave it unresolved any longer.

RhoSigma-QB64:
Looks good, except one critical issue I found some things you may want to cleanup in `qb64pe.bas`, the line numbers are according to your version of the file:

- stray comments in your native language in lines 2107, 2131, 2192 and 4119
- a stray space, align the `(LEN(` to the line above (now 25232)

Critical is the reset of `optionbase` in line 3040, this should go a couple lines down best into the variable block commented as `'reset altered variables` but at least below the `ide3:` label, because in the current position it's only executed when compiling from the command line, but not when compiling from the IDE. That said make sure you did compile tests from both. I mean compiling from command line obviously works, otherwise the GitHub tests would fail, so try if the tests which depend on the correct reset of the variable do also work as expected when compiling them from the IDE.

QB64Petr:
Well, exactly as I expected. I need to fix the err files. I'm going to do it.

RhoSigma-QB64:
See you've just removed the `optionbase = 0` but don't placed it with the other reset variables below the `ide3:` label.

Is that reset no longer needed?

QB64Petr:
@RhoSigma-QB64  I went through the code and confirmed that it is not needed there. I also checked it with tests, and nothing broke.

QB64Petr:
All requested changes, including the addition of ERR tests, have been completed. Please review them. Thank you.

RhoSigma-QB64 (approved):

SteveMcNeill (approved):

PR #701, opened 2026-03-22
Replace freeglut with GLFW

### LINKED ISSUE #128: $RESIZE:SMOOTH and STRETCH crop screens when maximized
As per the topic here: https://qb64phoenix.com/forum/showthread.php?tid=655

When the maximize button is used with $RESIZE:SMOOTH and $RESIZE:STRETCH, the screen overscales and clips part of the content, as it doesn't maintain the proper aspect ratio for the old screen resolutions.

### LINKED ISSUE #292: Race between `_FullScreen` and other screen-related commands
After the completion of #66 most commands that involve GLUT do not wait for the command to complete before returning, rather they simply add work to the GLUT command queue that will be done later on the GLUT thread at the next interval (typically 1/60 of a second). We don't bother waiting because many commands do not return anything or have any program-noticeable side-effects, so it's just not necessary.

This creates a problem with `_FullScreen` because although it really is a GLUT-based command it was not implemented by calling GLUT directly but rather by setting some global variables that the GLUT thread checks. This is a bit of an ugly way to do it, but also means that it happens independent of the GLUT command queue mentioned earlier, and some commands like `_SCREENMOVE` cause the program to exit fullscreen. Thus, sequences that should work like having `_SCREENMOVE _MIDDLE` before `_FULLSCREEN` actually don't, because the fullscreen can get handled before the screenmove (which then results in the screenmove exiting the fullscreen).

The solution here should be pretty simple, `sub__fullscreen` should not be setting global variables but instead should add a new 'fullscreen' command to the GLUT queue. That command will be executed in-order with all the other GLUT commands in the queue and things will work as intended.

### LINKED ISSUE #343: Extra Mouse Buttons.
My mouse has 3 extra mouse buttons then the _MOUSEBUTTON command can do, you could add a way to detect the pressing of extra buttons.

### LINKED ISSUE #597: IDE keeping keys pressed down after releasing on Linux.
It happened every time ANY key is held, the more you hold the more it's going to keep activated after release, it's very annoying trying to delete something only for the IDE to delete 8 more characters than you wanted.

Video illustrating the issue with scrolling:

https://github.com/user-attachments/assets/2f4c7dba-3fc2-4cbb-a1a9-0e32767e2c84

### LINKED ISSUE #608: Replace GLUT with something like GLFW / RGFW

a740g:
This PR represents a major architectural modernization of the QB64-PE graphics, windowing, timing, and input subsystems. It completely replaces the legacy FreeGLUT and GLEW backend libraries with GLFW 3.5 and GLAD, and resolves long-standing issues across Windows, macOS, and Linux.

New/improved functionality:

1. refreshRate& = `_DESKTOPREFRESHRATE` - Returns the vertical refresh rate (in Hz) of the monitor currently displaying the application window. If called before a window exists, it returns the refresh rate of the primary monitor.
2. `_WINDOWSIZELIMIT [(minWidth, minHeight)] [- (maxWidth, maxHeight)]` - Restricts the user-resizable dimensions of the window to minimum and/or maximum width and height bounds.
3. `_MOUSECURSOR imageHandle& [, (hotspotX&, hotspotY&)]` - Sets a custom hardware mouse cursor from a QB64-PE image handle.
4. seconds# = `_UPTIME` - Returns the application uptime as a high-resolution, monotonic double-precision floating-point number in seconds.
5. isDisabled& = `_MOUSEDISABLED` Returns _TRUE if the mouse cursor is currently in disabled/raw motion capture mode (via `_MOUSEHIDE _DISABLE`), or _FALSE otherwise.
6. handle%& = `_WINDOWHANDLE[(handleType&)]` - Extended with an optional handleType& argument to query specific native OS and graphics API handles:
      - 0 (or omitted): HWND on Windows, NSWindow* on macOS, Window XID on Linux X11.
      - 1: HDC on Windows, NSView* on macOS, Display* on Linux X11.
      - 2: HGLRC on Windows, NSOpenGLContext* on macOS, GLXContext on Linux X11.
      - 3: GLXWindow on Linux X11.
7. `_MOUSEHIDE [_DISABLE]` - Added optional _DISABLE parameter. When passed, the mouse cursor is hidden and locked in raw input mode, providing unlimited delta motion for 3D camera controls and FPS navigation without cursor hitting display edges. Works on Windows, Linux and macOS!
8. scroll# = `_MOUSEWHEEL[(axis&)]` - Added optional axis& parameter to support horizontal and multi-axis scrolling.
9. Sub-pixel precision for mouse & joystick coordinates - _MOUSEX, _MOUSEY, _MOUSEMOVEMENTX, _MOUSEMOVEMENTY, _MOUSEMOVE, _AXIS, _WHEEL have been upgraded to return DOUBLE for sub-pixel precision and better support for high-DPI displays.

In addition, substantial refactoring was done. `libqb.cpp` dropped from 31408 to 27185 lines.

Closes #608, #128, #292, #597, #343


grymmjack:
@a740g is this ready to test?

a740g:
> @a740g is this ready to test?

Yes, mostly. During testing, you might encounter a few quirks or incomplete features. Feel free to share your feedback here.

grymmjack:
> > @a740g is this ready to test?
> 
> Yes, mostly. During testing, you might encounter a few quirks or incomplete features. Feel free to share your feedback here.

Do we have a list of tests to run, or an automated test suite that runs for this kind of PR?

If not I can make a list of tests to run?

a740g:
> > > @a740g is this ready to test?
> > 
> > 
> > Yes, mostly. During testing, you might encounter a few quirks or incomplete features. Feel free to share your feedback here.
> 
> Do we have a list of tests to run, or an automated test suite that runs for this kind of PR?
> 
> If not I can make a list of tests to run?

We have a few that are passing under `tests/compile_tests/glut`. But we might need more. Since these changes are more visual and affect the user experience, I recommend doing manual checks (using the IDE is one way).

SteveMcNeill:
One thing I've noticed with the last artifact I was testing:  the keymappings need a little tweaking for when modifer keys are in effect.

Currently CTRL reports both uppercase and lowercase letters for input.  For example, depending on caps lock state, you may generate a CTRL-J, or you may generate a CTRL-j.   I've never seen any program in my days that tended to require CTRL-(check capslock)-keypress operability.

When CTRL or ALT is held, the key needs to always report the base key and not any lowercase/uppercase version.  We speak of things like CTRL-SHIFT-1 for a program, not CTRL-EXCLAMATION_MARK.

a740g:
> One thing I've noticed with the last artifact I was testing: the keymappings need a little tweaking for when modifer keys are in effect.
> 
> Currently CTRL reports both uppercase and lowercase letters for input. For example, depending on caps lock state, you may generate a CTRL-J, or you may generate a CTRL-j. I've never seen any program in my days that tended to require CTRL-(check capslock)-keypress operability.
> 
> When CTRL or ALT is held, the key needs to always report the base key and not any lowercase/uppercase version. We speak of things like CTRL-SHIFT-1 for a program, not CTRL-EXCLAMATION_MARK.

I've recovered an old (but fully functional) e-waste keyboard with an AltGr key from work. I'll eventually work on the AltGr stuff Rho mentioned on Discord, and while I'm at it, I'll check this out too. It might not happen this week, though, since I have a lot to take care of at home.

RhoSigma-QB64 (approved):

SteveMcNeill (approved):

PR #703, opened 2026-04-06
Nested static arrays in TYPE: revision fixesNested static arrays in TYPE: revision fixes

QB64Petr:
It addresses the problems raised in previous review rounds. In particular:
- `UCASE$` remains unchanged
- normal arrays and nested arrays use separate `Option Base` state
- the nested-array `LEN` size issue is fixed
- string member-array handling/documentation is included
- small cleanup was done in the touched paths

Relevant ERR tests added


QB64Petr:
Just a quick follow-up — everything should be ready on my side now and all checks are passing.

Since this is a fairly large change, I understand it needs proper review, but I’d appreciate any feedback or approval when possible so we can move this forward.

Let me know if there’s anything I should update.

a740g:
What happened to #699?

QB64Petr:
@a740g  Sorry for the confusion — this happened because I was creating a new branch for every update. I’m no longer doing that, as we discussed.

If that would be better, I can submit the changes to 699 and close 703. Would you agree?

a740g:
> @a740g Sorry for the confusion — this happened because I was creating a new branch for every update. I’m no longer doing that, as we discussed.
> 
> If that would be better, I can submit the changes to 699 and close 703. Would you agree?

If that is not too much trouble, then yes please. We also have Matt's review in #699.

PR #711, opened 2026-04-20
Fix: _Mem wasn’t handling the parent array in nested arrays

QB64Petr:
I found a bug in _Mem where it did not handle the parent array of nested arrays correctly. This fix resolves the issue, so _Mem now works with both the parent array and the nested array. I also added a new test to cover this case.


QB64Petr:
THIS IS BAD! I broke LEN. Damn. I'll do another test to compare LEN( () ) and _MEM.Size, it has to be the same.  I write repair now.

a740g (approved):

QB64Petr:
No, it can not be merged. LEN need repair.

a740g:
@QB64Petr I thought we already had tests for `LEN` right? If yes, all of those passed. How is `LEN` broken?

a740g:
Ok. I take that back. I do not see udt array tests for LEN.

QB64Petr:
@a740g Because I wasn't comparing array and element sizes with _MEM.Size, plus it was broken too. Size.bas is for normal arrays, not nested ones, I forgot to add that test. I'm working on a fix.

PR #715, opened 2026-04-22
_Preserve Fix

QB64Petr:
Added missing out-of-memory checks.
_Preserve behaves the same as before.
Added the _Retain function, which behaves correctly for both 1D and multidimensional arrays, unlike _Preserve.
Added IDE-side validation of declared array ranges. Hi, Steve :)
Added tests.


source/subs_functions/syntax_highlighter_list.bas
```diff
@@ -110,7 +110,7 @@ listOfKeywords$ = listOfKeywords$ +_
 
 ' [R] - Keywords alphabetical (1st line = QB64, 2nd line = QB4.5, 3rd line = OpenGL)
 listOfKeywords$ = listOfKeywords$ +_
-"_R2D@_R2G@_READBIT@_READFILE$@_RED@_RED32@_RESETBIT@_RESIZE@_RESIZEHEIGHT@_RESIZEWIDTH@_RGB@_RGB32@_RGBA@_RGBA32@_ROL@_ROR@_ROUND@" +_
+"_R2D@_R2G@_READBIT@_READFILE$@_RED@_RED32@_RESETBIT@_RESIZE@_RESIZEHEIGHT@_RETAIN@_RESIZEWIDTH@_RGB@_RGB32@_RGBA@_RGBA32@_ROL@_ROR@_ROUND@" +_
```
RhoSigma-QB64: Please keep alphabetic order here, _RETAIN should be after _RESIZEWIDTH, thanks.

QB64Petr: Oh! Repaired

QB64Petr:
I would like to add a few notes about this new version of _Preserve and _Retain.

The IDE now includes detection of array declarations using negative values (if Lbound > Ubound, error message occur). I later found out that the QB4.5 IDE does not catch this case, while our IDE now does. I am not completely sure whether this could conflict with the project/forum compatibility policy, so I would like this part to be reviewed.

I also tested _Preserve specifically because of its unusual historical behavior. The current _Preserve output is now consistent with the _Preserve behavior from IDE 4.3, which I used for comparison.

A new _Retain function has also been added. It is essentially similar to _Preserve, but with the expected and correct behavior for 1D and multidimensional arrays.

While working on this version, I also noticed an issue with Out of memory handling that I had missed in the previous version. This has now been fixed. I apologize for overlooking it earlier; at the time, I simply did not realize the problem was there.

For these reasons, a fairly large block of C code generated as a string was added to QB64PE.bas. Although I believe everything is in good order, I would still like to ask for thorough testing before allowing the merge.

SteveMcNeill (approved):
I see no issues with catching bugs which QB45 missed.  If someone could produce even one example, no matter how esoteric, to show where allowing the code to run *worked* in some odd and unusual manner, then I'd say we needed to keep it.  But this?  It's simply an error that occurs at runtime, which can be detected (partially) at design time.

What's the purpose of the IDE generating errors for us, except to catch them at design time to prevent them from causing issues at runtime.  I see no viable reason why you wouldn't catch these type of errors and report them to the user in the IDE.  Why waste their time making them compile, just to learn *then* that they made a glitch, when you can find it and report it to them while they're still in the coding process itself and not in the running/testing process which comes after compilation?

Issue #705, opened 2026-04-12
Undesired aspect ratio with DRAW TA


jessodum:
I think I count this issue as both a **bug** and a **feature request**.

The issue is with the TA subcommand in the DRAW statement (specifically using SCREEN 9, though I suspect it's an issue with all the original SCREEN modes, 1-13, as I'll discuss below).  There appears to be some undesirable "squishing" when there's rotation of a simple drawing.

First here is my test code:
```
DefInt A-Z
angle = 0
actScr = 1
visScr = 0

Screen 9
Print "Hit any key to start"

For i = 0 To 390 Step 30

    angle = i

    'Set SCREEN 9 (hi-res, hi-color graphics) specifying active and visible pages (buffers)
    Screen 9, , actScr, visScr
    Cls ' Clear (blank) the screen

    If angle > 360 Then angle = angle - 360

    ' Call DrawSquare subroutine
    DrawSquare angle

    ' Swap active and visible pages (buffers) for next iteration
    If actScr = 1 Then
        actScr = 0
        visScr = 1
    Else
        actScr = 1
        visScr = 0
    End If

    Sleep
Next

End

Sub DrawSquare (a)
    'DRAW Syntax:
    'BM200,200 ... Blind Move to 200, 200 (square center)
    'TAa ......... Turn Angle a
    'BR50 ........ Blind Right 50
    'BD50 ........ Blind Down 50
    'L100 ........ Left 100
    'U100 ........ Up 100
    'R100 ........ Right 100
    'D100 ........ Down 100
    'BM200,200 ... Blind Move to 200, 200 (square center)
    'P4,15 ....... Paint (fill 4=Red, border 15=White)
    '
    'Note: While the angle of the lines are correct, their length appears to be based
    '      on the some weird scaling factor.  At 0 degrees, the lengths are correct.
    '      But even at 90 degrees, it looks more like a rectangle than a square.
    Draw "BM200,200 TA" + Str$(a) + " BR50 BD50 L100 U100 R100 D100 BM200,200 P4,15"

    Locate 22 'Jump to text row 22
    Print "Current Angle: " + Str$(a)

End Sub
```

The results can be seen in the screenshots:

<img width="650" height="384" alt="Issue screenshot from original code at angle 0" src="https://github.com/user-attachments/assets/d5912ec7-6d0d-4c88-a49c-1b29b286824b" />
[...]

PR #717, opened 2026-04-24
Setting to allow force 1:1 aspect ratio

jessodum:
To resolve issue #705.

The "Compiler Settings" dialog in the IDE seemed to be the best fit to add this setting, even though it is a very specific use case flag.

I'm open to suggestions for a better implementation.

I'm not sure how the comment in `FUNCTION CompactMemberRefLayout$` (in `qb64pe.bas`) is changed.  Maybe some sort of encoding change from my editor?  If you can tell what actually changed and how to fix it (if needed), please let me know.


SteveMcNeill:
We've been discussing this and don't really seem to think that this needs to be something which is a compiler setting in the menu.  As of now, there's a couple of ideas for which would suit the language being discussed, but two of the options that seem to work best would be something like a ``DRAW _ForcedAspect "your draw commands like usual"``, or a change like we did with the sound statements with Play, just to expand it to something such as ``DRAW "F1 commands F0"``, whereas the F1/F0 would be a command built into the DRAW command itself to toggle this aspect ratio behavior.

RhoSigma-QB64:
Indeed this shouldn't be implemented as a general compiler option, as it just fixes one specific issue in on command.
As @SteveMcNeill already mentioned, this should be a fix to the affected command `DRAW` itself. best implemented as native draw command rather than a additional parameter `_ForceAspect` or similar. However, `F0/F1` as suggested by Steve is obviously not possible as that letter is already used for down/right diagonal lines. A better choice would be `S0/S1`, as that letter already deals with pixel sizes and it's current minimum value is 4, hence the values 0/1 would be possible for your specific use. In alternative take any other letter which isn't used yet by `DRAW`.

jessodum:
Thanks for the comments.  I'll put this in draft and pursue the alternatives (likely a new `DRAW` letter).

What about a modifier on `TA`?  I'll need to study the code to see if any other `DRAW` command letters are utilizing `r` and `ir`.

a740g:
> Thanks for the comments. I'll put this in draft and pursue the alternatives (likely a new `DRAW` letter).
> 
> What about a modifier on `TA`? I'll need to study the code to see if any other `DRAW` command letters are utilizing `r` and `ir`.

All the good letters are already taken. 😄 I am looking at the docs - using `TA`, `R` and `S` can be confusing IMO. `Q0` and `Q1` looks good to me. 🤷‍♂️

jessodum:
Heh. Every once in a while, I randomly think about how I should have gotten back to this so it wasn't just hanging out there.  Sorry about that.  Maybe one day I'll try again with the above ideas.

PR #722, opened 2026-04-27
Fix nested array parser validation and ERASE empty target

QB64Petr:
Parser repair for cases: Erase ()  and a().b(1)


QB64Petr:
So the syntax is working correctly for us now. But what if I changed the test where the syntax was wrong? Yes, I changed it. But in a different folder..... in my own folder yes, but not in the one that goes on github....

a740g (approved):

PR #728, opened 2026-05-06
Add QBJS Web Build option to Run menu.

boxgaming:
This feature was created in response to the following request on the QB64PE forum:

"What would be absolutely great?  To be able to use the QB64PE IDE and then have an option to "Export as QBJS and Run". 

Details and discussion of this feature can be found in this thread:
https://qb64phoenix.com/forum/showthread.php?tid=4598


RhoSigma-QB64:
The `qbjs-build.bas` code should better go into `internal\support\converter` and the respective code to compile and/or call it on demand be implemented as a new method in `source\ide\ide-converters.bas`. You may then call the method from your new build dialog as needed.

Also the use of `$IF..$ELSE..$ENDIF` precompiler blocks is not possible inside the QB64 sources itself, as all code must be compiled to get a new bootstrap `internal\source` snapshot working for all platforms. Hence this must be rewritten to use the `_OS$` function instead.

boxgaming:
> Also the use of `$IF..$ELSE..$ENDIF` precompiler blocks is not possible inside the QB64 sources itself, as all code must be compiled to get a new bootstrap `internal\source` snapshot working for all platforms. Hence this must be rewritten to use the `_OS$` function instead.

Ok, that shouldn't be a problem to change.

> The `qbjs-build.bas` code should better go into `internal\support\converter` and the respective code to compile and/or call it on demand be implemented as a new method in `source\ide\ide-converters.bas`. You may then call the method from your new build dialog as needed.

I can certainly move the qbjs-build source to the internal\support\converter directory.  The new build dialog is fairly interactive and most of the code there is UI-related.  There's not a single method that would be called from the UI, so I'm not sure I understand the utility of moving pieces of this to the ide-converters.bas file.

RhoSigma-QB64:
Ok, I see your point regarding the `ide_converters.bas` file. However, if you take a moment to analyze the methods in it you'll see that the converters will only be compiled once and the executables are saved in the `internal\support\converter` folder.

I.e. If the converter method is called, it will check if the exe already exists or if it first needs to be compiled. It's a matter of user experience to keep convert times as low as possible. In your dialog you ALWAYS compile your helper code. You may simply add the exe exist checks in your dialog if you don't want another method for it, but you really should avoid unnecessary recompiling of static code that doesn't change.

internal/support/converter/qbjs-build.bas
```diff
@@ -0,0 +1,440 @@
+$Console:Only
```
RhoSigma-QB64: Please convert this file to UPPER case keywords, it's our de-fecto standard for all BASIC sources.

boxgaming:
> In your dialog you ALWAYS compile your helper code. You may simply add the exe exist checks in your dialog if you don't want another method for it, but you really should avoid unnecessary recompiling of static code that doesn't change.

Actually, the existing code already checks for the presence of the qbjs-build binary and only compiles it if needed:
https://github.com/boxgaming/qb64pe/blob/b374d069b6072c7b7c8ed805935222dafec23930/source/ide/ide_methods.bas#L14310

internal/support/converter/qbjs-build.bas
boxgaming: OK

RhoSigma-QB64 (approved):

RhoSigma-QB64:
>Actually, the existing code already checks for the presence of the qbjs-build binary and only compiles it if needed.

Ok, didn't noticed that flag, but however, I still feel like the compiled build tool should rather go to `internal/support/converters` folder along with its source and be called from that location. It's cleaner than just writing it into _CWD$, and it's the way how all internal helper tools work.

Except from that it seems to be everything in green condition.

boxgaming:
> I still feel like the compiled build tool should rather go to `internal/support/converters` folder along with its source and be called from that location. It's cleaner than just writing it into _CWD$, and it's the way how all internal helper tools work.

Hrm, ok, I'll look into that...  I was thinking there would be utility in easily accessing the qbjs-build executable for direct command line usage (or from other build tools like Notepad++ or VSCode).  A number of the checks in the current logic assume that the build tool will be run from the same directory as the qb64 executable.

RhoSigma-QB64:
Well, your operations can still use the qb64 folder as root, it's just a matter of specifying the `internal/support/converter` path in the `-o` option when compiling the qbjs-build tool and later use the same path in the SHELL call. Shouldn't everything else be unaffect by that change?

boxgaming:
> Shouldn't everything else be unaffect by that change?

As it turns out there were a few changes needed, but not too bad.  I've submitted the requested updates.

RhoSigma-QB64:
Live-tested the whole thing now and got a "Path access error" (76) in line 440 (qbjs-build.bas). I've several spaces in the path, please look if you can reprouce the error on your system and take actions accordingly.

RhoSigma-QB64:
Here's the contents of .qbjs-build-out:

NodeJS Version:    v12.18.3
Checking current QBJS release version...
.....
QBJS Web Build:    v0.10.5-beta
Program Directory: C:\Users\Herzlich Willkommen\Downloads\qb64pe\internal\support\converter\
QB64 Directory:    C:\Users\Herzlich Willkommen\Downloads\qb64pe\
Source Directory:  C:\Users\Herzlich Willkommen\Downloads\qb64pe\ScreenBlankers\
Downloading QBJS v0.10.5-beta................ 200 
Download complete.
Unzipping QBJS...
Unzip complete.
Deleting zip.
Missing dependencies file:
C:\Users\Herzlich Willkommen\Downloads\qb64pe\internal\support\converter\.qbjs\qbjs-0.10.5-beta\export\dependencies.txt

boxgaming:
> Live-tested the whole thing now and got a "Path access error" (76) in line 440 (qbjs-build.bas). I've several spaces in the path, please look if you can reprouce the error on your system and take actions accordingly.

Interesting... I'll investigate.  I didn't see this issue on either the windows (11) or Linux (Mint) testing that I did (which included paths with spaces).  Curiously, the file paths reported in the .qbjs-build-out on my windows testing all have trailing backslashes whereas the ones you posted above do not.

I'll see if I can recreate the issue and add some additional logic to ensure a normalized path is used.

RhoSigma-QB64:
Ah, guess I see the reason, look at the last line (missing dependency...
>C:\Users\Herzlich Willkommen\Downloads\qb64pe\internal\support\converter.qbjs\qbjs-0.10.5-beta\export\dependencies.txt

shouldn't the `converter.qbjs` part not read `converter \ .qbjs` instead, so it's maybe just a missing seperator somewhere in the flow.

RhoSigma-QB64:
Forget the last comment, just tested. Damn GitHub strips the trailing backslashes when pasting the contents here and it also obviously strips a backslash when it is followed by a dot.

To make it clear, yes I've the trailing backslashes too in my file and it also reads correctly `converter \ .qbjs`, so that was all just GitHub messing with the pasted content.

RhoSigma-QB64:
Ok, sorry for the confusion I may have caused @boxgaming , it seems to be my outdated Win7 was the cause of all trouble. After taking a closer look into the qbjs-build.bas tool I saw you're using `tar` for extracting the QBJS archive, which is not available on Win7 per default. Hence the unpacking failed and that lead to the missing dependencies.txt file and the path access error. After I've downloaded and unpacked the QBJS archive manually using 7z the QBJS Web Build worked as advertised.

As we advertise Win7 as lowest supported version for QB64-PE, there should be a way to fall back to 7z or a respective error message in the build dialog should ask the user to manually unpack the downloaded archive with whatever tool he prefers. Of course, the delete of the download archive should be skipped in that case.

In alternative you could do it the same way as you already did for `Node.js` and redirect the user to a website where he can download a 3rd party version of `tar` first.

boxgaming:
Aaaah, ok, that makes sense now.  I’ll look into adding the logic to detect whether tar is available.

Out of curiosity, how long are you planning to support windows 7?  Microsoft ended support for it in 2020.

RhoSigma-QB64:
>Out of curiosity, how long are you planning to support windows 7? Microsoft ended support for it in 2020.

No idea, my wish and also best guess would be as long as possible, until changes on the C/C++ and/or Windows API level make it unavoidable to drop Win7 for the sake of advancing QB64-PE. But I'm rather seldom in touch with those things, the C/C++ stuff is rather maintained by @a740g and @mkilgore .

boxgaming:
I just decided to replace the shell call to "tar" with a powershell command since it is available on windows 7 and should work on all supported Windows versions.

a740g (approved):

RhoSigma-QB64:
>I just decided to replace the shell call to "tar" with a powershell command since it is available on windows 7 and should work on all supported Windows versions.

Unfortunately NO, the used method requires .NET 4.5, but Win7 came with .NET 3.5. Even if someone updated to 4.5 (like me), the PowerShell will not use it automatically, you've first to tweak the config how I learned now. So I started to try things by myself and the only method which really works "Out of the box" is using the standard shell ComObject, just replace the SHELL call with that below, it should also work for more recent Windows versions.

```
SHELL "powershell " + Q$("$shell = New-Object -ComObject Shell.Application; $zipFile = $shell.NameSpace(\" + CHR$(34) + qbjsParentDir + PathSeparator + releaseTag + ".zip\" + CHR$(34) + "); $destination = $shell.NameSpace(\" + CHR$(34) + qbjsParentDir + "\" + CHR$(34) + "); $destination.CopyHere($zipFile.Items(), 20)")
```

boxgaming:
Ah, good find.  I’ll try it out.  I’m also going to finish setting up a windows 7 virtual machine so I can more effectively test out the variations.

boxgaming:
Ok, I was able to get a windows 7 virtual machine environment setup and independently verify that your suggested changes work in both windows 7 and 11.

RhoSigma-QB64:
Ok, sent you an invitation as collaborator, when the checks have passed you should be able to merge.

PR #729, opened 2026-05-09
Runtime errors leads to the program locking up if `SUB _GL` is used.

QB64Petr:
This fix reapir  this bug:

SCREEN 0
_PRINTSTRING (0, 0), "error"

END

SUB _GL
END SUB

This bug is in Discord thread.
Now program continue as expected and crash not  

The current version on non-US Windows only downloads the 32-bit compiler. This version fixes this and gives you the choice on a 64-bit system whether you want the 32-bit or 64-bit version of the compiler.

### DIFF
--- a/internal/c/libqb.cpp
+++ b/internal/c/libqb.cpp
@@ -48,6 +48,7 @@
 
 // These are here because they are used in func__loadfont()
 #include <algorithm>
+#include <atomic>
 #include <string>
 #include <vector>
 
@@ -171,7 +172,7 @@ extern "C" int QB64_Resizable() {
     return ScreenResize;
 }
 
-int32 sub_gl_called = 0;
+thread_local int32 sub_gl_called = 0;
 
 static int32 image_qbicon16_handle;
 static int32 image_qbicon32_handle;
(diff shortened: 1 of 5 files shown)


internal/c/libqb.cpp
```diff
@@ -171,7 +172,7 @@ extern "C" int QB64_Resizable() {
     return ScreenResize;
 }
 
-int32 sub_gl_called = 0;
+thread_local int32 sub_gl_called = 0;
```
mkilgore: Why is this being made thread local?

internal/c/libqb.cpp
```diff
@@ -733,6 +734,49 @@ int64 display_lock_request = 0;
...
 
+extern uint8 close_program;
+extern uint8 suspend_program;
+extern uint8 stop_program;
+
+#ifdef DEPENDENCY_GL
+static std::atomic<int32> display_lock_modal_count{0};
+#endif
+
+void gui_modal_lock_begin() {
```
mkilgore, line 745: Are there supposed to be calls to these functions somewhere?

mkilgore (commented):
Could you share a link or better explain the issue we're trying to fix?

internal/c/libqb.cpp
QB64Petr: it is this:

https://discord.com/channels/975381912350752819/1183167369519046696/threads/1265653914628522037

Why is this being made thread local:

Because sub_gl_called is being used as an execution-context flag, not as shared program state. In GL builds, QBMAIN and SUB _GL run on different threads. When the flag is global, the QB thread can falsely see sub_gl_called = 1 just because the render thread is currently inside _GL, and then runtime code can think the QB thread is executing from _GL when it is not.

Making it thread_local preserves the intended check: END / SYSTEM from inside _GL still sees the flag on the GL thread, but the QB main thread no longer gets a false positive from the render thread.

mkilgore: > When the flag is global, the QB thread can falsely see sub_gl_called = 1 just because the render thread is currently inside _GL, and then runtime code can think the QB thread is executing from _GL when it is not.

In what situation does the QB thread call into `evnt()` again while `_GL` is running on the GLUT thread?

The QB thread should not be doing any active work while the _GL is called on the GLUT thread, that's why the 'display lock' exists along with the busy loop in `evnt()`. So while I think you're right that marking it `thread_local` doesn't break anything it shouldn't be fixing any bugs either.

Is marking it `thread_local` intended to fix something, or was it just added just because?

internal/c/libqb.cpp
QB64Petr: oh! I add missing updated files gui.h and gui.cpp!

internal/c/libqb.cpp
QB64Petr: You're right. After re-checking this, the thread_local sub_gl_called change is not needed for the modal dialog deadlock fix, and I do not have a separate reproducer proving that the global sub_gl_called causes this issue.

I'll remove that part from this PR and keep the fix focused on the modal dialog / display-lock deadlock.

internal/c/libqb.cpp
```diff
@@ -733,6 +734,49 @@ int64 display_lock_request = 0;
...
+    // dialog blocks the QB thread and would otherwise prevent evnt() from doing
+    // this acknowledgement.
+    if (display_lock_request > display_lock_confirmed)
+        display_lock_confirmed = display_lock_request;
+#endif
+}
+
+void gui_modal_lock_end() {
+#ifdef DEPENDENCY_GL
+    if (sub_gl_called)
```
mkilgore: Now that I'm seeing how it works I suspect this is why `thread_local` was on `sub_gl_called` - if `gui_modal_lock_end()` happens while the render thread is in _GL (which can happen, the main thread is blocked on the message box and can continue at any time) then this check will be true and the `display_lock_modal_count` will not be dropped, that breaks the synchronization going forward. The `thread_local` would have masked this issue since it made this check basically do nothing (this never gets run on the render thread running _GL).

I think we can safely remove the `sub_gl_called` checks from both the lock acquire and release functions - we still want to do this work even if _GL is currently running.

Edit: Took out the middle part about `lock_begin` that was wrong :D

QB64Petr: You're right. I removed the sub_gl_called checks from both gui_modal_lock_begin() and gui_modal_lock_end().

QB64Petr: @mkilgore  Is there anything else blocking approval?

mkilgore (approved):

PR #730, opened 2026-05-31
Fix division by zero diagnostics and UDT member case correction

QB64Petr:
This PR fixes three IDE/compiler issues:

1. Adds source line tracking for division-by-zero diagnostics, so internal crashes report the BASIC source line/module instead of only “Division by zero” or “Line: unknown”.

2. Fixes IDE case correction for indexed nested UDT member arrays, e.g.:
   Print Sin(f.z(1).xx(1))
   now preserves/corrects the member name according to the TYPE declaration:
   Xx

3. fixes an IDE crash caused by shared E being overwritten during nested UDT member reference parsing by using a local UDT element variable in udtreference$.

(try this source:)

CONST MAX_VERTICES = 15000
CONST MAX_FACES = 15000

TYPE Vector3D
    x AS SINGLE
    y AS SINGLE
    z AS SINGLE
END TYPE

TYPE Face3D
    v1 AS LONG
    v2 AS LONG
    v3 AS LONG
END TYPE

TYPE OBJModel
    vCount AS LONG
    fCount AS LONG
    vertices(1 TO MAX_VERTICES) AS Vector3D
    faces(1 TO MAX_FACES) AS Face3D
END TYPE

SUB LoadOBJ (filename AS STRING, model AS OBJModel)

model.vCount = 0
model.fCount = 0

DIM strip AS STRING
model.vertices(model.vCount).x = VAL(strip)
END SUB

Tested locally with a two-stage rebuild.


internal/c/libqb/src/error_handle.cpp
```diff
@@ -300,8 +323,32 @@ void error(int32_t error_number) {
     }
 
     // other critical errors
-    if (error_number == 11) {
-        gui_alert("Division by zero", "Critical Error", "ok");
+    if (error_number == QB_ERROR_DIVISION_BY_ZERO) {
+        char errmess[512];
```
a740g: We can use an `std::string` here.

a740g (approved):

internal/c/libqb/src/error_handle.cpp
QB64Petr: Yes, std::string is better. I rewrite it.

a740g: Nice!

a740g (approved):

Issue #559, opened 2024-10-17
memory-leak / corruption with large string assignments


mdijkens:
**Describe the bug**
When (re)assigning a string variable with a large string (>1GB) after a few times program crashes

**To Reproduce**
c$=space$(1123123123)
c$=space$(1123123123)
c$=space$(1123123123)
c$=space$(1123123123)
c$=space$(1123123123)
c$=space$(1123123123)
c$=space$(1123123123)

**Expected behavior**
Program ends/continues normally

**Screenshots**
Program just aborts/killed

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Version 3.14.1

**Additional context**
Ran into issue with c$=ReadFile$() reading multiple large files in a loop

mdijkens:
See also https://qb64phoenix.com/forum/showthread.php?tid=3130

mkilgore:
The underlying issue is already reported here: https://github.com/QB64-Phoenix-Edition/QB64pe/issues/278

But note that due to how the string buffer is handled you'll need almost twice the total memory for the strings you create (so if you want to make 32GB of strings, you need around 64GB of total memory). Really, to be able to handle that much data the `qbs` string logic needs to be significantly improved.

Issue #278, opened 2022-12-21
qbs_data_size can wrap when large strings are created, which leads to the program crashing.


mkilgore:
`qbs_data` is basically a big buffer used to back all of the `qbs` strings. It's unclear to me why it's used vs. just separate allocations for each buffer, but I haven't dug into it super closely. The issue here is that `qbs_data_size` is a `uint32`, even on 64-bit systems. Since that variable controls the size of `qbs_data`, if we attempt to create enough strings large enough that they would cause `qbs_data_size` to be larger than 4GBs (the max uint32 value), then it wraps back to zero. We then call `realloc(qbs_data, qbs_data_size)` with the now way too small `qbs_data_size` and effectively just free most of the memory backing the strings in the program, leading the program to crash soon afterward.

The fix here is to use the proper type `size_t` for these array size variables `qbs_data_size` and `qbs_sp`. `size_t` is always the proper size to hold the max array size, so it won't overflow. Additionally we should add a check for the overflow condition and error the program with an OOM error (the overflow could still happen in a 32-bit application if you attempt to allocate way too much memory).

Raven-Singularity:
This seems like a pretty bad bug, memory corruption and crashing when allocating large amounts of data.

No updates?

PR #744, opened 2026-07-08
Fix qbs large string arena and $ErrorLocation handling

QB64Petr:
This change fixes large-string memory handling in the qbs runtime and adds explicit runtime source-location tracking through `$ErrorLocation`.

The qbs string data arena now uses platform-sized counters for arena size and stack position, and its growth paths check for overflow before reallocating. This prevents wrapped arena sizes from corrupting memory or crashing programs when creating, concatenating, or assigning very large strings.

The qbs arena growth logic now:

* detects arithmetic overflow before allocation,
* reports Out of memory instead of continuing with wrapped sizes,
* preserves existing live qbs data if realloc fails,
* retries with the exact required size when preferred geometric growth fails,
* correctly rebases live qbs string pointers when realloc moves the arena,
* preserves binary-safe string operations, including strings containing embedded NUL bytes.

This addresses the large-string arena issues tracked by #278 and #559.

A native regression/stress test was added for the qbs issue #278 path. It covers arena counter sizes, overflow boundaries, negative allocation, large qbs_set paths, binary-safe concatenation, realloc relocation, fragmentation compaction, realloc failure, exact-size retry behavior, qbs_set overflow handling, and configurable large-arena stress testing.

This change also adds `$ErrorLocation:On` / `$ErrorLocation:Off`.

`$ErrorLocation` is a whole-program diagnostic option. It is off by default and is independent from `$Checking`. When enabled, the generated program tracks the current BASIC source line before executing statements so fatal runtime errors that bypass the normal error-event path can still report the BASIC source location.

The fatal-error reporting path now uses the tracked BASIC location for:

* critical Out of memory errors,
* Division by zero,
* Out of stack space,
* dynamic library load failures,
* missing dynamic-library SUB/FUNCTION symbols,
* `_GL` calls outside `SUB _GL`,
* `END` / `SYSTEM` called inside `SUB _GL`.

Dynamic library loading and symbol lookup run during generated initialization / registration code, outside the normal statement path. `$ErrorLocation:On` now records the corresponding DECLARE line only when such a load or lookup failure occurs, avoiding stale source locations after successful loads.

Fatal shutdown now routes through `libqb_exit()`. When a GLUT/OpenGL window is active, exit handling is transferred to the GLUT-owning thread; if the current thread already owns GLUT, it exits directly instead of queueing a message to itself. This avoids deadlock during fatal-error shutdown from OpenGL-related paths.

Test 1:

$ErrorLocation:On

OPTION _EXPLICIT

SCREEN _NEWIMAGE(320, 200, 32)

DO
    _LIMIT 60
LOOP

SUB _GL
    SYSTEM
END SUB

Test 2:

$Console:Only
$ErrorLocation:Off

Option _Explicit

Print "Starting recursive stack test with source tracking disabled."
RecurseStack 0
System

Sub RecurseStack (depth As Long)
    RecurseStack depth + 1
End Sub

Test 3:

$Console:Only
$ErrorLocation:On

Option _Explicit

Print "Starting recursive stack test with source tracking enabled."
RecurseStack 0
System

Sub RecurseStack (depth As Long)
    RecurseStack depth + 1
End Sub

and also test from your Github:

c$=space$(1123123123)
c$=space$(1123123123)
c$=space$(1123123123)
c$=space$(1123123123)
c$=space$(1123123123)
c$=space$(1123123123)
c$=space$(1123123123)

program normal continue in 64bit OS, Out of memory occur in 32bit OS.

### DIFF
--- a/internal/c/libqb/src/error_handle.cpp
+++ b/internal/c/libqb/src/error_handle.cpp
@@ -6,13 +6,18 @@
 #include <stdlib.h>
 #include <string.h>
 
-#include <string>
+#ifdef QB64_WINDOWS
+#include <windows.h>
+#elif defined(QB64_LINUX) || defined(QB64_MACOSX)
+#include <pthread.h>
+#endif
 
 #include "command.h"
 #include "error_handle.h"
 #include "logging.h"
 #include "event.h"
 #include "gui.h"
+#include "glut-thread.h"
 
 uint32_t new_error;
 uint32_t error_occurred;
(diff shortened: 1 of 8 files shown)


source/qb64pe.bas
```diff
@@ -1685,6 +1688,11 @@ SetPreLET "_DEBUG_", _TOSTR$(GetRCStateVar(vWatchOn))
 'auto-includes beforefirstline.bi and afterlastline.bm to include/exclude
 'network specific code
 SetPreLET "_SOCKETS_", _TOSTR$(GetRCStateVar(SockDepOn))
+IF GetRCStateVar(ErrorLocationMode) = 2 THEN
```
RhoSigma-QB64, line 1691: Would you please explain what use cases you see by exposing the $ErrorLocation state as predefined variable for the pre-compiler?
```vb
$IF _ERRORLOCATION_ THEN
    'What useful things could be done here depending on the state of this?
$END IF
```

source/qb64pe.bas
```diff
@@ -14393,10 +14476,10 @@ FUNCTION ParseCMDLineArgs$ ()
                 CMDLineSwitch = _TRUE
```
RhoSigma-QB64, line 14477: Please @QB64Petr , you must always make sure to update your local repo to the the latest updates from the main repo before doing changes and PRs, you partly revert some of the changes I did lately here and it could have been easily overseen when it was longer ago than just a couple days.

source/qb64pe.bas
```diff
@@ -5718,6 +5771,15 @@ DO
             WriteBufLine MainTxtBuf, "sf_mem_lock=mem_lock_tmp;"
             WriteBufLine MainTxtBuf, "sf_mem_lock->type=3;"
 
+            ' Native recursion can exhaust the C++ thread stack before error(256)
+            ' is reached. Check every user SUB/FUNCTION entry while enough stack
+            ' remains to report the error. $ErrorLocation controls only whether
+            ' the report includes a BASIC source position; stack exhaustion must
+            ' still be caught when source tracking is disabled. The runtime helper
+            ' uses platform-specific stack bounds behind one common interface.
+            WriteBufLine MainTxtBuf, "extern bool error_check_stack();"
```
mkilgore, line 5780: We should expose it via a header and a `libqb_` call, rather than define an `extern` here.

internal/c/libqb/src/glut-main-thread.cpp
```diff
@@ -113,6 +113,9 @@ static void initialize_glut(int argc, char **argv) {
 }
 
 static bool glut_is_started;
+// True only on the initial thread that owns GLUT. thread_local avoids any
+// platform-specific thread-ID API and works on Windows, Linux and macOS.
+static thread_local bool current_thread_owns_glut;
```
a740g, line 118: Most of this is going to change in GLFW PR. FYI

source/qb64pe.bas
```diff
@@ -5718,6 +5771,15 @@ DO
...
+            WriteBufLine MainTxtBuf, "if (error_check_stack()) goto exit_subfunc;"
```
mkilgore, line 5781: Why are we jumping to the end of the sub/function rather than just exiting the program after reporting the error?

tests/c/qbs_issue278_tests.cpp
```diff
@@ -0,0 +1,511 @@
...
+    reset_runtime();
+
+    const char left_bytes[] = {'A', '\0', 'B', 'C'};
+    const char right_bytes[] = {'D', 'E', '\0', 'F', 'G'};
+    qbs *joined = qbs_add(qbs_new_txt_len(left_bytes, sizeof(left_bytes)), qbs_new_txt_len(right_bytes, sizeof(right_bytes)));
+
+    const char expected[] = {'A', '\0', 'B', 'C', 'D', 'E', '\0', 'F', 'G'};
+    REQUIRE(joined != nullptr);
+    REQUIRE(joined->len == static_cast<int32_t>(sizeof(expected)));
+    REQUIRE(std::memcmp(joined->chr, expected, sizeof(expected)) == 0);
```
mkilgore, line 230: Could you switch these tests to use `test.h` (how `buffer.cpp` and `http.cpp` currently work)? You can add to `test.h` if you need something that it doesn't have, but in general it'll produce more consistent results and gives a bit more information about the test results (Ex. `test_assert_buffers()` can compare these buffers and will display the buffers themselves if they don't match).

We also need to wire them up to run on build [here](https://github.com/QB64-Phoenix-Edition/QB64pe/blob/990a6c788e1401585dd3bee2639238020aa1812c/tests/build.mk#L13) and [here](https://github.com/QB64-Phoenix-Edition/QB64pe/blob/main/tests/run_c_tests.sh#L7).

internal/c/libqb/src/error_handle.cpp
```diff
@@ -28,15 +33,119 @@ static uint32_t inclercl;
...
+    }
+
+    srcLine = 0;
+    srcFile = NULL;
+    return false;
+}
+
+// Formats fatal runtime errors without dynamic allocation. Static storage
+// keeps the out-of-stack path from allocating another object while reporting
+// the original failure.
```
mkilgore, line 73: This is fine but I would note `gui_alert` does allocations, so we should keep in mind we're not really avoiding them.

source/qb64pe.bas
QB64Petr: You're absolutely right. This is my mess, I'll rewrite it to the current version. I am sorry.

source/qb64pe.bas
QB64Petr: Yeah. It's there because I sent the wrong version from development.

internal/c/libqb/src/glut-main-thread.cpp
QB64Petr: Okay. Thanks for the information. So wouldn't it be better if I wait? In the meantime, I can address other reported issues with dynamic nested arrays in qb64pe.bas

source/qb64pe.bas
QB64Petr: If the program had terminated immediately right here, it would have been simpler, but less correct. It would have broken the unified runtime error model and might have bypassed cleanup. This is simply "controlled escape", not "overlooking the error".

mkilgore: Are you picturing people trying to catch the stack overflow error and continuing? I don't think that's a good idea to support personally, by skipping over all the code in the current SUB/FUNCTION we're impacting the program in a very unpredictable way even if people are intentionally trying to catch it.

FWIW QB doesn't allow catching the error either, so there's no compatibility concerns to consider.

internal/c/libqb/src/glut-main-thread.cpp
a740g: I'll think it's ok. Looks simple. We can go ahead. I'll include the functionality in my branch.

PR #746, opened 2026-07-11
fix: widen arraydesc to LONG and disarm include-open error trap (>32,767 identifiers crash)

### LINKED ISSUE #745: Compiler crashes on programs with >32,767 identifiers; crash is masked as a bogus "File REDIM... not found" error
**Version:** QB64-PE v4.5.0 (Windows 11 x64, self-built and release binary both affected)

## Summary

Two independent bugs in `source/qb64pe.bas` combine to make large programs fail with a nonsensical error:

1. **`arraydesc` is a 16-bit INTEGER but holds a LONG identifier index.** Once a program registers more than 32,767 identifiers, a `REDIM` of an *already-declared* array whose id exceeds 32,767 silently wraps `arraydesc` negative, and the compiler crashes internally with Error 9 (Subscript out of range) at the `cmemlist(arraydesc)` lookup.
2. **The `$INCLUDE` open error-trap flag is never disarmed on success**, so the internal crash from bug 1 (or any other internal compiler error) is caught by the global error handler, resumed into the include-open code path, and reported as `File <internal element string> not found` — e.g. `File REDIMARR(1TO20)ASSTRING not found`.

The result is an error message that names no real problem, points at a syntactically valid line, and appears/disappears with unrelated source edits (anything that shifts identifier registration order moves which array crosses id 32,768). This cost us multiple days of bisecting a 190K-LOC project before instrumenting the compiler itself.

## Minimal reproduction

Generate a program with 33,000 dummy identifiers plus a `ReDim` of an existing array (Python generator, or any equivalent):

```python
lines = ['Dim Shared filler%d As Long' % i for i in range(1, 33001)]
lines += ['ReDim Shared arr(1 To 10) As String',
          'ReDim arr(1 To 20) As String',
          'Print "ok"']
open('repro_ids.bas', 'w', newline='\r\n').write('\n'.join(lines))
```

```
qb64pe -z -q repro_ids.bas
```

**Observed (v4.5.0):**

```
File REDIMARR(1TO20)ASSTRING not found
Caused by (or after):REDIM ARR ( 1 TO 20 ) AS STRING
LINE 33002:ReDim arr(1 To 20) As String
```

**Expected:** successful compile (the program is valid; with the patch below it compiles and runs).

Note the reported "file name" is the compiler's internal element string for the statement being translated. When the failing statement sits inside a `$INCLUDE`d file, the message becomes `File REDIM... not found in line N of <includefile> included`, which is even more misleading.

## Root cause 1 — id truncation (the crash)

`source/qb64pe.bas` line 623 (v4.5.0):

```basic
DIM SHARED arraydesc AS INTEGER
```

`arraydesc` is assigned `currentid` (declared `AS LONG`, line 640) in the `dim2` array-handling paths (`arraydesc = currentid: scope2$ = scope$`, many sites around lines 15495–16567). The id tables (`ids`, `cmemlist`, `arrayelementslist`) are LONG-indexed and grown correctly in `regid` — only this variable truncates.

- First declaration of an array takes the `arraydesc = 0` branch and uses LONG `idn` — always fine. That's why the *declaring* `DIM`/`ReDim` compiles and a later `ReDim` of the same array crashes.
- For ids 32,768–65,535 the INTEGER wraps **negative** → Error 9 at line ~14264: `IF cmemlist(arraydesc) THEN cmem = 1`.
- For ids ≥65,536 it would wrap to a small **positive** value → no crash, but a lookup of the *wrong* array's `cmemlist` entry (silent misbehavior).

Confirmed by instrumenting the global error handler: `ERR=9, Subscript out of range` at the `cmemlist(arraydesc)` line while translating the second `ReDim`.

### Fix

```basic
DIM SHARED arraydesc AS LONG
```

A quick audit for other explicitly-INTEGER variables that hold identifier indexes is advisable (the file is `DEFLNG A-Z`, so only explicit `AS INTEGER` declarations are at risk).

## Root cause 2 — error masking (the bogus message)

Both `$INCLUDE` managers arm an ON-ERROR trap before opening an include file:

```basic
qberrorhappened = -2   ' (or -3 in the first manager)
OPEN f$ FOR BINARY AS #fh
qberrorhappened2:
IF qberrorhappened = -2 THEN
    ...
    EXIT FOR           ' <-- success path exits with the flag still armed
END IF
qberrorhappened = 0    ' <-- only reached on the failure path
```

On a **successful** open, `EXIT FOR` skips the reset, and the check just below (`IF qberrorhappened <> -2 THEN ... "File " + a$ + " not found"`) passes *because* the flag is still `-2` — after which it is never cleared. The stale `-2` persists for the rest of compilation.

When any later internal compiler error fires, the global error handler sees `qberrorhappenedvalue = -2` and does `RESUME qberrorhappened2`. Control falls through the include-open block (the handler already set `qberrorhappened = 1`), reaches the failure branch, and emits `File <a$> not found` — where `a$` is whatever scratch string the translator was holding (here, the element form of the statement being translated).

### Fix

Disarm the trap immediately after a successful open, in both managers:

```basic
        IF qberrorhappened <> -2 THEN
            qberrorhappened = 0: a$ = "File " + a$ + " not found": GOTO errmes
        END IF
        qberrorhappened = 0  ' disarm on success, so later internal errors surface honestly
        inclevel = inclevel + 1: ...
```

With this fix alone (without fix 1), the repro reports an honest `UNEXPECTED INTERNAL COMPILER ERROR!`-class failure instead of the garbage filename, which would have made bug 1 findable in minutes.

## Verification

With both fixes applied to v4.5.0 source and the compiler rebuilt:

- The 33k-identifier repro compiles and runs.
- Our 190K-LOC / ~3,500-routine project (which hit this daily, with the failing line moving around as we edited unrelated code) compiles cleanly and passes its full test suite (42/42).

PR with both fixes: incoming (will link below). Instrumentation traces available on request.

bharvey42:
Fixes #745.

## What

Two small fixes in `source/qb64pe.bas` for compiling large programs:

1. **`arraydesc` INTEGER → LONG.** `arraydesc` holds `currentid` (a LONG index into the identifier tables). Once a program registers more than 32,767 identifiers, a `REDIM` of an already-declared array whose id exceeds 32,767 wrapped `arraydesc` negative and crashed the compiler with Error 9 at the `cmemlist(arraydesc)` lookup. For ids ≥65,536 the wrap goes positive instead — no crash, but a silent read of the wrong array's `cmemlist` entry.

2. **Disarm the `$INCLUDE` open error trap on success (both managers).** The success path (`EXIT FOR`) left `qberrorhappened` armed at `-2`/`-3` for the remainder of compilation, so any later internal compiler error resumed into the include-open code and was reported as `File <internal element string> not found` (e.g. `File REDIMARR(1TO20)ASSTRING not found`) instead of surfacing as a real error. This is what made bug 1 so hard to diagnose.

## Repro / verification

Minimal repro from #745 (33,000 dummy `Dim Shared` lines + a `ReDim` of an existing array):

- Unpatched v4.5.0: `File REDIMARR(1TO20)ASSTRING not found`
- With this patch: compiles and runs.

Also verified on a real-world 190K-LOC / ~3,500-routine project that hit this daily: with a patched compiler it builds cleanly and passes its full test suite (42/42). The failing statement previously moved around with unrelated source edits, since anything shifting identifier registration order changes which array crosses id 32,768.

With fix 2 alone, the repro reports an honest internal-error message instead of the bogus filename, which makes any future bug of this class diagnosable.


source/qb64pe.bas
```diff
@@ -634,7 +634,7 @@ cleariddata.dynudtargmode = ""
 DIM SHARED gosubid AS LONG
 DIM SHARED redimoption AS INTEGER
 DIM SHARED dimoption AS INTEGER
-DIM SHARED arraydesc AS INTEGER
+DIM SHARED arraydesc AS LONG 'holds currentid (LONG); as INTEGER it wrapped negative once a program registered >32767 ids
```
mkilgore, line 637: The comment about "AS INTEGER..." is unnecessary.

source/qb64pe.bas
```diff
@@ -3128,6 +3128,7 @@ DO
             qberrorhappened = 0
         NEXT
         IF qberrorhappened <> -3 THEN qberrorhappened = 0: a$ = "File " + a$ + " not found": GOTO errmes
+        qberrorhappened = 0 'disarm the include-open error trap after success, else later internal errors resume here and report as a bogus "File ... not found"
```
RhoSigma-QB64, line 3132: No comment here either, just write `qberrorhappened = 0` like all the other occurrences of it. (both managers)

RhoSigma-QB64 (approved):

PR #747, opened 2026-07-11
feat: widen the UDT index field from 9 to 12 bits (raises the silent 511-TYPE cap, adds honest error)

bharvey42:
Related: #745 (same discovery session; independent change — this does not depend on #746).

## Problem

User-defined TYPEs are silently capped at **511**. The low bit-field of an encoded type id (`AND 511`) holds either a primitive's size in bits or, when `ISUDT` is set, the UDT index. The `udtx*` storage tables grow automatically (`increaseUDTArrays`), so nothing stops you *defining* TYPE #512 — but every `AS Type512` reference encodes as `ISUDT + 512`, and `512 AND 511 = 0` aliases it to the wrong slot. Depending on what lives there, this surfaces as "Element not defined", "Invalid expression", or **silent wrong-type codegen**, with no diagnostic pointing at the real cause. Large data-driven programs do hit this (ours has ~500 TYPEs across 190K LOC).

## Why this is a small change

All `IS*` flags start at bit 20 (`ISOFFSET = 2^20`), and bits 9–19 of the type word are **unused** — the index field can widen without moving a single flag constant. Primitive sizes max out at 256 bits (`FLOATTYPE`), so a wider extraction mask is value-neutral for non-UDT types.

## What this PR does

- Adds `CONST UDTMASK = 4095` (12 bits → 4,095 TYPEs; documented headroom to `&HFFFFF` if ever needed) in `source/utilities/type.bi`.
- Replaces every low-field extraction (`AND 511`, plus the composite masks like `AND (ISFLOAT + ISUDT + 511 + ...)`) with `UDTMASK` across `qb64pe.bas`, `utilities/type.bas`, and `ide/ide_methods.bas` — 280 occurrences, byte-level replace, no other content touched.
- Adds an explicit compile error — `Too many user-defined types (maximum 4095)` — at TYPE registration, so exceeding the (new) cap is an honest diagnostic instead of silent aliasing.

Write sites need no change: type words are built as `ISUDT + index`, which already carries indexes up to 2^20 uncorrupted; only the read masks truncated. The runtime's own `& 511` sites in `libqb.cpp` all operate on `qbs_input_variabletypes`, which never carries UDT-indexed words (UDTs cannot be `INPUT`/`READ` targets), so no runtime change is needed.

## Verification

- Generated torture program: 600 TYPEs, with types #512+ exercised through scalars, nested UDTs, static and dynamic arrays (incl. `REDIM _PRESERVE`), member arrays with `LBOUND`/`UBOUND`, fixed-string members, and `_MEM`/`_MEMGET`. Unpatched compiler: fails with `Element not defined` at the first `v512.a` reference. Patched compiler: compiles; all runtime assertions pass. Happy to attach the generator script.
- A real-world 190K-LOC / ~500-TYPE / ~3,500-routine project compiles cleanly with the patched compiler and passes its full test suite (42/42).

Functionally the same patch was validated on v4.5.0 sources; this branch applies it to current `main` (the idiom is unchanged, occurrence count differs).

### DIFF
--- a/source/ide/ide_methods.bas
+++ b/source/ide/ide_methods.bas
@@ -9102,7 +9102,7 @@ FUNCTION idevariablewatchbox$ (currentScope$, filter$, selectVar, returnAction)
                         IF ok = -2 THEN
                             getid usedVariableList(tempIndex&).id
                             IF id.t = 0 THEN
-                                typ = id.arraytype AND 511
+                                typ = id.arraytype AND UDTMASK
                                 IF id.arraytype AND ISINCONVENTIONALMEMORY THEN
                                     typ = typ - ISINCONVENTIONALMEMORY
                                 END IF
@@ -9634,7 +9634,7 @@ FUNCTION idevariablewatchbox$ (currentScope$, filter$, selectVar, returnAction)
                             usedVariableList(varDlgList(y).index).elementOffset = ""
                             getid usedVariableList(varDlgList(y).index).id
                             IF id.t = 0 THEN
-                                typ = id.arraytype AND 511
+                                typ = id.arraytype AND UDTMASK
                                 IF id.arraytype AND ISINCONVENTIONALMEMORY THEN
                                     typ = typ - ISINCONVENTIONALMEMORY
                                 END IF
--- a/source/qb64pe.bas
+++ b/source/qb64pe.bas
@@ -2278,12 +2278,12 @@ DO
 
                                 'Calculate element's size
                                 IF typ AND ISUDT THEN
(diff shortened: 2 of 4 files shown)


RhoSigma-QB64:
Was wondering about the connection to PR #512 (Migrate to buffered search/recent files histories (Problem/Verification sections), but obviously this is caused by GitHubs hashtagging and your use of 512 in the text, maybe you should edit this to not trigger the hashtagging and let the 512 describe what you really intended.

RhoSigma-QB64 (approved):

PR #751, opened 2026-07-19
Dynamic nested arrays fixes, $ErrorLocation and more

QB64Petr:
This PR contains:
- Dynamic nested arrays fixes (critical bug repaired)
- Now is possible using static or dynamic nested arrays as sub/function parameter
- Added $ErrorLocation (works for Division by zero, Out of memory, Out of stack space, Dynamic library if not found, Dynamic  library - if sub or function in dynamic library not exists, _GL command if is called outside of _GL's scope, END/SYSTEM called within SUB _GL 's scope). $ErrorLocation in default is disabled, so programs speed is the same as in old versions.
-  Repairs dynamic nested arrays compilation speed
-  fix for selecting source code text by dragging the mouse  (no longer need to move the mouse sideways if you are highlighting source code and the mouse moves outside the IDE window)
- fix error messages in IDE for dynamic nested fields


source/utilities/type.bas
```diff
@@ -825,13 +826,22 @@ SUB increaseUDTArrays
 END SUB
 
 ' Build the alternate physical layouts used by nested member arrays.
-' Layout mode 1 is the forced-only descriptor layout for explicit _Dynamic
-' members under otherwise inline storage. Layout mode 2 is the full dynamic
-' layout used by REDIM/runtime descriptor paths. Static legacy layout remains
-' stored in udtxsize/udtesize and is not modified here.
+' Layout mode 1 is the canonical descriptor layout for explicit _Dynamic members
+' when the TYPE graph has no runtime-bound member arrays. Layout mode 2 is the
+' canonical layout for TYPE graphs that do contain runtime-bound member arrays.
```
mkilgore: What is a "runtime-bound" member array?

source/utilities/type.bas
```diff
@@ -1019,7 +1056,7 @@ FUNCTION UDTDynMembersOK% (udt_index AS LONG, layout_mode AS LONG)
                     UDTDynMembersOK% = 0
                     EXIT FUNCTION
                 END IF
-                Give_Error "Dynamic TYPE layout supports only fixed-length strings for now"
+                Give_Error "This TYPE combines dynamic nested arrays with an inline variable-length STRING array; declare that STRING array with _Dynamic or use STRING * length"
```
mkilgore: Why is this not allowed?

source/utilities/type.bas
```diff
@@ -844,12 +854,32 @@ SUB BuildUDTDynLayoutM (udt_index AS LONG, layout_mode AS LONG)
...
     DO WHILE member_id
+        memberalign = 8
+        IF UDTMemberDynDesc%(member_id, layout_mode) THEN
+            memberalign = ptrbits
+        ELSEIF (udtetype(member_id) AND ISUDT) <> 0 THEN
+            nested_udt = udtetype(member_id) AND UDTMASK
+            IF UDTDynHasMemberArrays%(nested_udt, layout_mode) THEN memberalign = ptrbits
+        END IF
+
+        ' Descriptor slots are dereferenced as ptrszint** in generated C. Align
```
mkilgore, line 846: This doesn't really seem necessary given we don't align anything in a TYPE currently.

QB64Petr: These are old dead branches of code that I want to remove in the future. The functionality is the same as before.

source/utilities/type.bas
QB64Petr: I meant:

a member array in a TYPE whose declared bounds are evaluated only at runtime

in English, approximately runtime-bound member array.

I'll remove this too. In the next release. The point here is that there is still a branch that originally ran (mode 2) in the case where there was not a numeric value in TYPE, but a variable. And this branch is no longer used today, because that there will be no variable in TYPE. There will always be a number or a constant. But if I were to cook it out now, it would be for a month, maybe longer. And since you want to release as soon as possible, I left it there.

RhoSigma-QB64 (approved):

source/utilities/type.bas
mkilgore: Are you talking something like this?

```
l& = 20
TYPE foo
    bar(l&) As Long
END TYPE
```

IMO that should never be supported, so that makes it easy 🤣 But if that's the case then why are we able to trigger the errors about mismatched layout in the current version? Considering that the functionality was never actually implemented anyway I would think it should be impossible to have layout=2, but I hit the errors from dim/redim differences.

source/utilities/type.bas
mkilgore: Sorry, are you saying the alignment logic is something you want to remove in the future? If that's the case then why are you adding it in this PR?

source/qb64pe.bas
```diff
@@ -14622,6 +14637,15 @@ FUNCTION UDTRuntimeMemberArrays% (udt AS LONG)
 END FUNCTION
 
 
+'Select the one physical descriptor-layout mode allowed for a TYPE graph.
+'This is declaration-driven: call sites never upgrade or rewrite the mode.
+SUB GetUDTCanonMode (udt AS LONG, mode_out AS LONG)
```
mkilgore: Could this be a FUNCTION?

tests/compile_tests/arrays/type_dynamic_string_slot_32bit.bas
```diff
@@ -0,0 +1,14 @@
+$Console:Only
```
mkilgore, line 1: We need to add at least one test similar to [t220_varstring_scalar_around_dynfield](https://github.com/QB64-Phoenix-Edition/QB64pe/blob/main/tests/compile_tests/arrays/t220_varstring_scalar_around_dynfield.bas) but declaring the TYPE as a non-array. That's the case that was blowing up for me and it doesn't appear to be directly covered by these tests.

QB64Petr: @mkilgore  you think something as this?

$Console:Only
$Unstable:TypeFields

Option _Explicit

Type T220
    headText As String
    nums(0 TO 1) _Dynamic As Long
    tailText As String
End Type

Dim ok As _Byte
ReDim work As T220

work.headText = "alpha"
work.nums(1) = 456
work.tailText = "omega"
ok = (work.headText = "alpha" And work.nums(1) = 456 And work.tailText = "omega")
If ok Then Print "PASS t220_varstring_scalar_around_dynfield" Else Print "FAIL t220_varstring_scalar_around_dynfield"
System

This test run correctly. Or this:

$Console:Only
$Unstable:TypeFields

Option _Explicit

Type T220
    headText As String
    nums As Long
    tailText As String
End Type

Dim ok As _Byte
ReDim work As T220

work.headText = "alpha"
work.nums = 456
work.tailText = "omega"
ok = (work.headText = "alpha" And work.nums = 456 And work.tailText = "omega")
If ok Then Print "PASS t220_varstring_scalar_around_dynfield" Else Print "FAIL t220_varstring_scalar_around_dynfield"
System

this also run correctly. OR this:

$Console:Only
$Unstable:TypeFields

Option _Explicit

Type T220
    headText As String
    nums As Long
    tailText As String
End Type

Dim ok As _Byte
ReDim work(20) As T220

work(20).headText = "alpha"
work(20).nums = 456
work(20).tailText = "omega"
ok = (work(20).headText = "alpha" And work(20).nums = 456 And work(20).tailText = "omega")
If ok Then Print "PASS t220_varstring_scalar_around_dynfield" Else Print "FAIL t220_varstring_scalar_around_dynfield"
System

this is also correct.

mkilgore: Yes your first test is what I mean, you should make that into a testcase.

source/utilities/type.bas
QB64Petr: Sorry, my previous reply was unclear. By “old dead branches” I did not mean the new alignment logic.

I meant the old call-site-driven layout mode inference code, where the compiler discovered the required dynamic UDT layout while processing calls, updated the SUB/FUNCTION argument mode, and then requested another full compilation pass. That is the code path I would like to remove once the declaration-driven canonical layout is fully in place.

The alignment logic serves a different purpose.

The descriptor layout contains actual pointer-sized descriptor slots. In the generated C++ code, those slots are accessed as pointer-sized values, currently through casts such as ptrszint **. For that access to be well-defined and portable, the slot address must have its natural pointer alignment. Without alignment, a descriptor slot could begin at an arbitrary byte offset after a small member such as _BYTE or INTEGER.

We would otherwise need to replace every direct descriptor load and store with an unaligned-safe implementation, for example using memcpy. Keeping the slots naturally aligned allows the existing direct pointer access to remain simple and avoids architecture-dependent unaligned dereferences.

This does not change the legacy packed layout of ordinary QB64PE TYPEs. The padding is added only to the internal descriptor layout used for dynamic nested arrays. The beginning of nested descriptor layouts is aligned as well, and the final layout size is padded so that the descriptor slots in the next element of a UDT array remain aligned.

The alignment was added during the same redesign as the canonical layout mode, but it is not required because recompilation was removed. Removing the repeated recompilation made the physical descriptor layout declaration-driven and stable, which meant that its offsets and array stride had to be defined consistently. The alignment logic ensures that this stable layout can be safely accessed by the generated C++ code.

mkilgore: I understand the concern but we already do that for all the existing TYPE members today (cast them to a pointer to their type and read them, regardless of their alignment). It seems odd to me to add padding around these members (and then at the end of the TYPE to ensure array's align) when we don't do that for anything else.

source/qb64pe.bas
```diff
@@ -5720,6 +5740,15 @@ DO
...
             WriteBufLine MainTxtBuf, "sf_mem_lock->type=3;"
 
+            ' Native recursion can exhaust the C++ thread stack before error(256)
+            ' is reached. Check every user SUB/FUNCTION entry while enough stack
+            ' remains to report the error. $ErrorLocation controls only whether
+            ' the report includes a BASIC source position; stack exhaustion must
+            ' still be caught when source tracking is disabled. The runtime helper
+            ' uses platform-specific stack bounds behind one common interface.
+            WriteBufLine MainTxtBuf, "extern bool error_check_stack();"
+            WriteBufLine MainTxtBuf, "if (error_check_stack()) goto exit_subfunc;"
```
mkilgore: Sorry, I didn't realize this never got addressed, my comments from the last PR are still the same:

- The `error_check_stack()` should be a `libqb_` function and exposed via a header rather than an inserted `extern`
- I don't think we should allow handing the stack overflow via jumping to the end of the function, it should just exit the program after displaying the error. I'm open to thoughts from others on it but 🤷 very few languages allow handling a stack overflow because of how problematic it is.

source/utilities/type.bas
```diff
@@ -1009,17 +980,16 @@ FUNCTION UDTDynMembersOK% (udt_index AS LONG, layout_mode AS LONG)
...
+        ' owner-aware lifecycle path so its qbs* slot is initialized, copied and freed safely.
         ELSEIF (udtetype(member_id) AND ISSTRING) <> 0 THEN
             IF (udtetype(member_id) AND ISFIXEDLENGTH) = 0 THEN
                 IF UDTDynOwnerOK%(udt_index, layout_mode) THEN EXIT FUNCTION
                 IF Error_Happened THEN
                     UDTDynMembersOK% = 0
                     EXIT FUNCTION
                 END IF
-                Give_Error "Dynamic TYPE layout supports only fixed-length strings for now"
+                Give_Error "This TYPE combines descriptor-backed nested arrays with an unsupported inline variable-length STRING array; declare that STRING array with _Dynamic or use STRING * length"
```
mkilgore: Please add a testcase showing how you hit this error.

source/utilities/type.bas
```diff
@@ -989,12 +960,12 @@ FUNCTION UDTDynMembersOK% (udt_index AS LONG, layout_mode AS LONG)
...
-                ' the recursive owner helpers. Keep that path explicit so ordinary inline
-                ' UDT arrays remain compatible with legacy layout.
-                IF UDTMemberDynDesc%(member_id, layout_mode) AND udtxvariable(nested_udt) THEN
+                ' An explicit _Dynamic UDT member array whose element owns variable strings
+                ' needs recursive owner helpers. An unmarked or _Static member array remains
+                ' inline; only its nested element layout becomes descriptor-aware when required.
+                IF UDTMemberDynDesc%(member_id) AND udtxvariable(nested_udt) THEN
                     IF udtearrayfieldmode(member_id) <> 2 THEN
-                        Give_Error "Dynamic TYPE UDT member arrays containing variable-length strings require _Dynamic for now"
+                        Give_Error "A TYPE array member whose element type contains variable-length STRING members must be declared with _Dynamic"
```
mkilgore: Please add a testcase showing how you hit this error.

source/utilities/type.bas
```diff
@@ -963,19 +934,19 @@ FUNCTION UDTDynMembersOK% (udt_index AS LONG, layout_mode AS LONG)
...
-                    ' descriptor-backed, they must be explicitly marked _Dynamic so
-                    ' legacy inline arrays do not silently change storage model.
-                    IF UDTMemberDynDesc%(member_id, layout_mode) THEN
+                    ' A variable-length STRING member array owns qbs* slots. Explicit _Dynamic
+                    ' stores those slots behind a live descriptor; unmarked and _Static members
+                    ' stay inline. The storage form is never inferred from DIM, REDIM or a call site.
+                    IF UDTMemberDynDesc%(member_id) THEN
                         IF udtearrayfieldmode(member_id) <> 2 THEN
-                            Give_Error "Dynamic TYPE variable-length STRING member arrays require _Dynamic for now"
+                            Give_Error "A variable-length STRING array inside a TYPE must be declared with _Dynamic"
```
mkilgore: Please add a testcase showing how you hit this error.

QB64Petr: These are remnants of the old branch. That's why I'm sending a patch to type.bas now, these states cannot be invoked.

QB64Petr:
The previous update remove uncallable error messages and this update fixes out of stack space. Here is a manual test for this (you need to re-compile qb64pe.bas first)

$CONSOLE:ONLY
$ERRORLOCATION:ON

' Manual runtime regression test for libqb_check_stack().
'
' Expected result:
'   1. This source compiles and links successfully.
'   2. Recursive entry eventually displays a fatal "Out of stack space" error.
'   3. The program terminates immediately after the fatal error.
'   4. The ON ERROR handler is not entered.
'   5. No line beginning with "FAIL:" is printed.
'
' This is intentionally a manual runtime test rather than a normal .output
' compile test, because success terminates the process through the fatal-error
' path and displays the platform's critical-error report.

Declare Sub ConsumeStack (depth As _Unsigned Long)

ON ERROR GOTO ErrorHandler

PRINT "Stack overflow fatal-error test started."
PRINT "Expected: Critical Error - Out of stack space"
PRINT

ConsumeStack 0

PRINT "FAIL: recursive SUB returned to the main program."
SYSTEM

ErrorHandler:
PRINT "FAIL: ON ERROR caught the stack overflow. ERR ="; ERR
SYSTEM

SUB ConsumeStack (depth AS _UNSIGNED LONG)
    ConsumeStack depth + 1

    ' This statement deliberately remains after the recursive call so the
    ' generated native call cannot be optimized into tail recursion.
    PRINT "FAIL: recursive call returned at depth"; depth
    SYSTEM
END SUB

QB64Petr:
I have no idea what's going on with the sound in Linux. I haven't done anything with it. Any ideas?

source/qb64pe.bas
a740g: I agree with @mkilgore 

We shouldn't allow recovery via `goto exit_subfunc`. this may leave the caller or the program in an inconsistent state. We should rather just show a critical runtime error and exit so that the author can re-design their code to work better.

a740g:
> I have no idea what's going on with the sound in Linux. I haven't done anything with it. Any ideas?

Looks weird to me. Try re-running.

QB64Petr:
> > I have no idea what's going on with the sound in Linux. I haven't done anything with it. Any ideas?
> 
> Looks weird to me. Try re-running.


tests/compile_tests/audiosoundtest/sound_test.bas

Compilation: 36: audio_sound_test/sound_test: result:
FAILURE! - Result is wrong:
EXPECTED:

'My Bon-nie lies O-ver the O-cean My Bon-nie lies O-ver the sea My Bon-nie lies O-ver the O-cean Oh bring back my Bon-nie to me..! My Bon-nie lies O-ver the O-cean My Bon-nie lies O-ver the sea My Bon-nie lies O-ver the O-cean Oh bring back my Bon-nie to me..! Foreground playback always takes longer to complete than background playback.' 

GOTH:
'My Bon-nie lies O-ver the O-cean My Bon-nie lies O-ver the sea My Bon-nie lies O-ver the O-cean Oh bring back my Bon-nie to me..! My Bon-nie lies O-Ver the O-cean My Bon-nie lies O-ver the sea My Bon-nie lies O-ver the O-cean Oh bring back my Bon-nie to me..!'

The third line is missing from that audio test - and it is time-dependent in the test:

if d2 - d1 > 5 Then
Print "Foreground playback always takes longer to complete than background playback."
End If

That means that in Linux it can be done earlier. So I go to start my Linux laptop and I have to check it.

QB64Petr:
@a740g 

So I ran the test in Linux Mint and it works fine even with the third line. So I wonder if the environment on the server is waiting for the playback to finish? I don't know what I could do about it. Everything works fine from the QB64PE perspective.

QB64Petr:
I believe this version addresses the concerns raised so far, including the original stack-related issue we discussed yesterday.

I tried to take all of the feedback into account and make the requested adjustments. The full test suite is passing now as well.

Please have a look whenever you have the time. There is no rush, and I would appreciate any further feedback.

source/utilities/type.bas
QB64Petr: old code, already removed

source/qb64pe.bas
QB64Petr: see if the latest patch meets your requirements

source/qb64pe.bas
QB64Petr: now it is function

source/utilities/type.bas
QB64Petr: I also removed this - it was a remnant of the old code

source/utilities/type.bas
QB64Petr: I also removed this - it was a remnant of the old code

QB64Petr:
I'm not merging it yet so you can review the latest fixes. I'll merge it when you approve.

mkilgore (approved):

RhoSigma-QB64 (approved):

Issue #753, opened 2026-07-21
Declare Sub/Func lines are not formatted


RhoSigma-QB64:
**Describe the bug**
`Declare Sub` and `Declare Function` outside of `Declare Library` blocks are not formatted according to the IDE settings (Options > Code layout...).

Although those old style `Declare` lines are no longer required in QB64(PE) and will be silently ignored by the Compiler they should nevertheless correctly be handled by the code formatter if those appear in the loaded/pasted code.

**To Reproduce**
Pasting the code below into the IDE (or running it through the commandline formatter) will not change it according to the current/given layout settings:
```qb
DecLarE   SuB         MySubRoutine(x aS InteGer)
DecLarE FunCtiON    MyFunction%(x   aS    InteGer)
```

**Expected behavior**
Correctly layout the code, e.g. for automatic single spacing and UPPER keywords.
```qb
DECLARE SUB MySubRoutine (x AS INTEGER)
DECLARE FUNCTION MyFunction% (x AS INTEGER)
```

**Versions**
All versions since ancient SDL ones, seems this was never thought of to be implemented.

PR #755, opened 2026-07-28
Format DECLARE SUB nad DECLARE FUNCTION in code layout 

QB64Petr:
repair for issue #753 (Declare SUB, Declare FUNCTION layout formating)


source/qb64pe.bas
```diff
@@ -5261,6 +5261,20 @@ DO
                 GOTO finishednonexec
             END IF
 
+            IF secondelement$ = "SUB" THEN
+                l$ = SCase$("Declare" + sp + "Sub")
+                IF n >= 3 THEN l$ = l$ + sp + getelements$(ca$, 3, n)
+                layoutdone = 1: IF LEN(layout$) THEN layout$ = layout$ + sp + l$ ELSE layout$ = l$
```
mkilgore, line 5267: Does this completely fix #753? I don't see how the types and such will the proper casing applied to them. It probably quickly gets more complicated though as you would have to parse the actual DECLARE format to know if something is a type name vs. a variable name.

Separately, we can probably add a `format_test` for this.

RhoSigma-QB64: Matt is right, tested the artifact and it definitly doesn't formats the parameters/types given to the SUB or FUNCTION. So this needs more adjustments, Probably the formatting behavior when defining SUB/FUNC can be cloned here or at least taken to find the right path to do it.

QB64Petr: Yeah. Exactly as you described it. I got a bit carried away (I was focusing only on the capitalization of DECLARE SUB / DECLARE FUNCTION), but... as you pointed out, I completely overlooked this. :)

So now, besides the issues already described, I also need to go back and find out how QB4.5 behaves when the SUB name in DECLARE SUB has different parameters, or a different number of parameters, than the actual SUB definition later in the code. The same applies to functions.

I'd say that this nice, quick fix has turned into something horribly more complicated. But since I've already started working on it, I'll keep going.

Basically, support for it will be added, but actually using it won't be required.

RhoSigma-QB64: Yes, but don't overdo it here with QB4.5 compatiblity. These DECLARE lines for SUB/FUNC have never had any relevance for QB64 and were always just completly ignored by the compiler. Hence there must not be any syntax check on it, just format it if they appear in old code and be done with it, the compiler can ignore those lines as it always did.

RhoSigma-QB64: And if it takes to much time right now, then simply close the PR now and reopen it later when there's more time for it. This is really not a high priority thing which must be solved as quick as possible. I just noticed it when adding Matt's QB64pe-Json library to the Libraries Pack, that those lines are not formatted. Matt obviously just used it in the top of his .bi file to document the API.

QB64Petr: Yes, I'll close this PR. The incomplete support would be confusing. For example:

Declare Sub Test (Value as Integer)
.
.
.
Sub Test (Text as String, Value as Long)
End Sub

And the IDE wouldn't catch this. That would be a real mess.

Issue #723, opened 2026-04-28
PRINT bug in console mode


CookieSnippets:
[https://qb64phoenix.com/forum/showthread.php?tid=4641](https://qb64phoenix.com/forum/showthread.php?tid=4641)

I consider this a severe (aka compatibility breaking) bug!!

```
$CONSOLE:ONLY
A$ = "hello" + CHR$(0) + "world"
PRINT "var: "; A$
PRINT "raw: hello" + CHR$(0) + "world"
SLEEP
SYSTEM
```
Expected output:
Just as I get in QB45, I expect this to be: Hello World

```
var: hello world
raw: hello world
```

What I do get instead is just: Hello

```
var: hello
raw: hello
```

=>

PRINT does not seem to handle strings with null characters in 'console mode'.
These should not be C-style strings, but BSTR...... which can include null characters!!

Is IS properly printed with QB45 in the very same console + other low ascii chars do work 'fine' (as far as they are printable)!
So, this is obviously the string which gets terminated because of the null char.

In graphics mode, it DOES work. There I do get the expected 'Hello World'.

---------------

Note: I havent tested other PRINT related stuff yet (PRINT USING, etc)

Win10 x64, QB64 4.4.0 (I suspect this is no different in 4.5, haven't tested yet, sorry)

PR #756, opened 2026-07-28
Fix console PRINT with embedded null characters (issue #723)

QB64Petr:
Fixes an issue with printing chr$(0) in the console, test is attached in tests/compile_tests/console_only


internal/c/libqb.cpp
```diff
@@ -11753,20 +11753,18 @@ void qbs_print(qbs *str, int32 finish_on_new_line) {
...
     if (write_page->console) {
-        static qbs *strz;
-        if (!strz)
-            strz = qbs_new(0, 0);
-        qbs_set(strz, qbs_add(str, qbs_new_txt_len("\0", 1)));
-        if (finish_on_new_line)
-            std::cout << (char *)strz->chr << std::endl;
-        else
-            std::cout << (char *)strz->chr;
+     if (str->len)
```
mkilgore: The change looks fine, the indentation is wrong.

tests/compile_tests/console_only/print_console_chr$(0).bas
```diff
@@ -0,0 +1,13 @@
+$Console:Only
```
mkilgore, line 1: Let's not put a '$' in the test file name.

internal/c/libqb.cpp
QB64Petr: I will correct both objections, the output file will be without spaces between words, because the image capture method on github removes null characters, so the text then appears to be without spaces.

tests/compile_tests/console_only/print_console_chr$(0).bas
QB64Petr: Test and output file renamed

internal/c/libqb.cpp
mkilgore: FWIW I downloaded the raw file from GitHub and there are no NUL bytes in it:
image

QB64Petr: CHR$(0) is emitted by the fixed QB64PE program as a real 0x00 byte. However, the compile test runner does not compare the program output byte-for-byte. It stores stdout in a Bash variable using command substitution:

testResult=$( ... "$EXE" ... 2>&1)

Bash variables cannot contain embedded NUL bytes. During command substitution Bash discards NUL bytes from the command output; since Bash 4.4 it may also report warning: command substitution: ignored null byte in input.

Therefore, if the executable writes:

68 65 6c 6c 6f 00 77 6f 72 6c 64
h  e  l  l  o     w  o  r  l  d

then after:

testResult=$(./program)

the Bash variable contains:

helloworld

not hello\0world and not hello world.

The QB64PE test runner also reads the .output file through command substitution:

expectedResult="$(cat ".../$testName.output")"

so this test infrastructure cannot preserve NUL bytes on either side of the comparison. It is a text comparison after Bash has removed NULs. The test still catches this particular regression: before the fix the console output stops at NUL and Bash receives hello; after the fix the executable outputs hello\0world, Bash removes only the NUL, and the comparison sees helloworld.

mkilgore: We could probably just fix the test runner, we already write each test result [to a file](https://github.com/QB64-Phoenix-Edition/QB64pe/blob/f73300c2df53e97566f7fbc7936af96149233dcf/tests/compile_tests.sh#L144) so we could pipe the result directly there instead of storing it in `testResult`.

`expectedResult` is similarly read from a file, we could skip that and just compare the two files with [`diff` instead of `==`](https://github.com/QB64-Phoenix-Edition/QB64pe/blob/f73300c2df53e97566f7fbc7936af96149233dcf/tests/compile_tests.sh#L149), and then remove the `show_incorrect_results` on the next line since we'll be able to see the `diff` output anyway.

tests/compile_tests/console_only/print_console_chr$(0).bas
mkilgore: Thanks, please also make sure to remove the '$' ones :)

internal/c/libqb.cpp
```diff
@@ -11753,20 +11753,18 @@ void qbs_print(qbs *str, int32 finish_on_new_line) {
     static uint32 character;
 
     if (write_page->console) {
-        static qbs *strz;
-        if (!strz)
-            strz = qbs_new(0, 0);
-        qbs_set(strz, qbs_add(str, qbs_new_txt_len("\0", 1)));
+        if (str->len)
+            std::cout.write((char *)str->chr, str->len);
```
a740g, line 11757: I would use fwrite() and fputc() for the leanest solution. But this works too. 👍

mkilgore: @a740g I partially agree but we'd really want to covert all the code off of `cout` to do that, there's a couple other spots scattered about.

tests/compile_tests.sh
```diff
@@ -56,7 +102,6 @@ do
     if [ "$OS" == "win" ]; then
         EXE="$EXE.exe"
     fi
-
```
mkilgore, line 59: Is there a reason you removed so many blank lines?

tests/compile_tests.sh
```diff
@@ -32,6 +32,53 @@ show_incorrect_result()
     printf "GOT:      '%s'\n" "$2"
 }
 
+trimmed_output_size()
```
mkilgore: Is the purpose of this to ignore any trailing newlines? Perhaps we could call it `filesize_without_trailing_newline()`.

Separately, is it not possible to fix the test files with trailing newlines? Or is there just a lot of them and it's too much of a pain?

QB64Petr: Yes. The purpose is to preserve the previous behavior of the test runner regarding trailing newlines.

Previously both the program output and the .output file were read through Bash command substitution. That implicitly discarded trailing newlines, but it also made the comparison unable to handle embedded NUL bytes.

Switching to file-to-file comparison fixed the NUL issue, but made trailing line endings significant and caused existing tests to fail even though they passed with the old runner.

trimmed_output_size() is therefore intended to reproduce that old newline-insensitive behavior while keeping the comparison binary-safe. It handles trailing LF/CRLF rather than removing NULs or otherwise treating the files as text.

Regarding fixing the .output files instead: technically we could normalize them, but that would change the existing test-suite convention and potentially require touching many unrelated expected-output files. I preferred to keep the historical behavior of the runner and make the NUL fix independent of a large test-data cleanup.

A separate normalization of the expected-output files could certainly be done later if that is preferred.


Regarding the function name, I agree that trimmed_output_size() is somewhat vague. Since the function specifically computes the file size while ignoring trailing line endings, a more descriptive name would be better.

I would slightly prefer filesize_without_trailing_newlines() over filesize_without_trailing_newline(), because the function may ignore more than one trailing newline and also needs to account for line-ending sequences such as CRLF. The plural makes the intended behavior a little clearer.

If you prefer filesize_without_trailing_newline(), though, I am fine with renaming it that way as well.

The removed blank lines were accidental formatting noise from my edit, not part of the intended change.

I write update.

mkilgore (approved):

a740g (approved):

PR #757, opened 2026-07-29
Small QoL improvements

a740g:
This PR adds several small QoL improvements. Most of these were in my GLFW branch.

- Adds clangd‑related local files to .gitignore
- Fixes typos and applies linting improvements to README.md
- Updates setup_lnx.sh with GLFW dependency handling (same as .github/workflows/build-process.yml)
- Ensures all PowerShell calls use -NoProfile
- Extends setup_win.cmd with support for building via MinGW available in the system PATH
- Adds OS autodetection in Makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,9 +2,23 @@
 # Disable implicit rules
 MAKEFLAGS += --no-builtin-rules
 
-ifndef OS
-$(error "OS must be set to 'lnx', 'win', or 'osx'")
+ifeq ($(filter-out Windows_NT,$(OS)),)
+# Attempt to auto-detect the OS
+	ifeq ($(OS),Windows_NT)
+		OS := win
+	else
+		OS_RAW := $(shell uname -s)
+		ifeq ($(OS_RAW),Darwin)
+			OS := osx
+		else ifeq ($(OS_RAW),Linux)
+			OS := lnx
+		else
+			$(error "OS must be set to 'lnx', 'win', or 'osx'")
+		endif
+	endif
 endif
+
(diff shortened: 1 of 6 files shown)


RhoSigma-QB64:
TBH, I don't understand that hype to detect and do everything automatically, where is the problem passing some things as parameters or simply asking the user. You should think about, that our lowest supported system is Windows 7 SP 1. That means not only QB64-PE itself must run on it, but also all setup scipts and makefiles. The repository's `setup_win.bat` definitly does not run on my Win7 since eons, there is no `curl` by default, no `tar` and most of the PowerShell crap doesn't work either. I can only grab the current repo builds from the artifacts. You should orientate on the oldest supported system not on the newest.

This is FYI only, I'm fine with using the artifacts. How many other users are on Win7 is ???, but I felt this had to be said.
Go ahead....

Makefile
```diff
@@ -2,10 +2,24 @@
 # Disable implicit rules
 MAKEFLAGS += --no-builtin-rules
 
-ifndef OS
-$(error "OS must be set to 'lnx', 'win', or 'osx'")
+ifeq ($(filter-out Windows_NT,$(OS)),)
+# Attempt to auto-detect the OS
+	ifeq ($(OS),Windows_NT)
```
mkilgore, line 7: MinGW already sets `$(OS)` by default? Guess I should have picked something else 🤣

mkilgore (approved):

Makefile
a740g: It's actually Windows :)

Windows Env Vars

PR #758, opened 2026-08-03
Fix whole-array assignment for all array types

QB64Petr:
fix for copying all types of arrays using the equals sign.

Now not only the zero index is copied, but all indexes correctly, it works for classic arrays and nested arrays. The worst thing was the placement of protections, so many ERR tests are added. The tests for this work are in the folder tests/compile_tests_arrays/ and they are files with names from t001 to t044.


QB64Petr:
There is a typo in the title of the last PR, it shouldn't be the word frixed but fixed, I didn't realize the line shift after inserting $console:Only. I hope that's all :)

QB64Petr:
The latest PR also includes a fix for dynamic UDT fields used in GET and PUT operations.

Overall, the PR adds support for whole-array assignment using the = operator. Arrays can be copied between a nested UDT array field and a regular array of the same element type—for example, from a nested LONG array to a regular LONG array and vice versa. It also supports copying between UDT arrays of the same type containing nested arrays, as well as copying between regular arrays.

The PR also adds tests for writing array contents to a file and reading them back, for both regular arrays and UDT arrays:

/tests/compile_tests/arrays/put_get_write_read_fixed_string
/tests/compile_tests/arrays/put_get_write_read_UDT_fixed_string

Whole-array assignment using the = operator is covered by tests t001????????? through t044???????????? in:

/tests/compile_tests/arrays/

Furthermore, this PR fixes a bug that allowed an array to be passed to a FUNCTION or SUB parameter that expected a scalar value. In that case, the FUNCTION or SUB incorrectly treated element zero of the array as the scalar argument.

This is covered by:

/tests/compile_tests/arrays/t045_function_scalar_array
/tests/compile_tests/arrays/t046_sub_scalar_array

source/qb64pe.bas
```diff
@@ -11876,6 +11895,20 @@ DO
                             END IF
                         END IF
 
+                        'A user SUB scalar parameter must not silently accept array() as array(0).
+                        'The original spelling is required because both forms currently encode offset zero.
+                        IF id2.internal_subfunc = 0 AND id2.ccall = 0 THEN
```
mkilgore: Is there not a more general solution to this? It's a problem in more locations than just specifically SUB/FUNCTION parameters.

a740g (approved):

PR #760, opened 2026-08-21
Add #line to C++ source for debugging

mkilgore:
The debug information generated by the C++ compiler (g++ or clang++) includes information that maps the generated machine code back to source lines. However, this information currently maps back to the generated C++ code, which is not very useful for most people.

This change has QB64-PE start generating `#line` entries before each generated C++ line for the user's code, each one will look like `#line "foobar.bas" 13`. This tells the C++ compiler that in the debug information it should map that C++ code (and it's coresponding machine code) to show as coming from `foobar.bas` at line `13`. The end result is that if you run your code in a debugger like `gdb` or `lldb` you now get a result like this, where the debugger will actually show you the original QB64 source line on a break:
image
In the same way, it also allows setting breakpoints at lines in your .bas source, `step` will go by .bas line rather than C++ line, etc.

The logic to add the `#line` is fairly simple and hooks into the existing writes to the buffers (with some care to handle the 'raw' writes with no newline, since `#line` has to be on its own line to work). To do this I did have to replace every write for `MainTxtBuf` and `DataTxtBuf`, so this change does touch a number of lines in `qb64pe.bas`, but the transformation itself is simple.

There's also some complexity with how the paths to the files are stored in the debug information, there's a new 'Compile Settings' option related to it. By default we place relative paths into the debug information, so you get `test.bas`, `foobar/test.bas`, `../outer.bas`, etc. For `internal` and `libraries` files they are listed relative to those folders, so they look like `internal/support/include/aftermain.bas`. This does work, but requires ensuring the debugger has the correct source file locations loaded for the source files to be found.

The alternative option is 'absolute paths', where the full path to the file on your machine is embedded into the debug information. This makes debugging on your machine easier, but is not as portable since if you move the folder or go on a different machine then the paths are no longer correct. Overall it's a generally minor issue though and relative paths are the default. If/when we have an integrated debugger based on `gdb`/`lldb` it would make sure the source paths are loaded correctly, so for most users it won't matter either way.


a740g (approved):
Love to see this replace VWATCH in the IDE. 😊

QB64Petr (approved):

source/qb64pe.bas
```diff
@@ -28159,6 +28173,89 @@ FUNCTION CompareVersions (v$, v1$)
...
+    incName = ResolveDebugPath$(incName, isInternal)
+
+    'line 0 is out of range for the C preprocessor, which we can end up with
+    'before the first line of the program has been read
+    IF lineNum < 1 THEN lineNum = 1
+
+    WriteBufLine buf, "#line " + _TOSTR$(lineNum) + " " + CHR$(34) + GetEscapedPath$(incName) + CHR$(34)
+END SUB
+
+SUB WriteBufLineCpp (buf As Long, s as String)
```
RhoSigma-QB64: Wouldn't it be better to place these two routines into `utilities/s-buffer/sb_qb64pe_extension.bm` as those are more ore less glue routines between QB64pe and the buffer system as the other routines in that file?

mkilgore: Yeah I can do that.

RhoSigma-QB64 (approved):

PR #761, opened 2026-08-22
Add QB64PE_NOPROMPT

mkilgore:
This give a way to tell the runtime that unhandled errors should not bring up an interactive prompt. The output text appears the same as with the prompt, but we do not wait for input and instead use the indicated action:

QB64PE_NOPROMPT=y
QB64PE_NOPROMPT=Continue

Both trigger their corresponding options on the dialog, 'y' exits the program on an unhandled error, 'continue' lets the program continue past it to the next line.

Tests will normally run with `y` so that unhandled errors cannot hang the test runner. I also added logic to the test scripts where a .noprompt file lets a tests configure the value (so it can test the 'continue' option).

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -28,6 +28,13 @@ sure the test output ends with the last meaningful line.
 Tests can also have an optional `*.flags` file. The contents of this file will
 be provide as command line arguments to QB64-PE when compiling the test source.
 
+The tests run with `QB64PE_NOPROMPT=y` set, this ensures that any unhandled or
+critical errors do not prompt and wait for input (which would hang the test
+runner). To change this behavior tests can have an optional `*.noprompt` file,
+the contents of this file will be provided as the value of `QB64PE_NOPROMPT`
+when running the test executable. This is mostly useful for testing the
+'Continue' behavior.
+
 Tests are compiled from the directory the `.bas` file is in. That is, they are
 compiled as though the user `cd`'d into the directory the `.bas` file is in,
 and then ran the compiler from that starting location to compile the `.bas`. If
--- a/internal/c/libqb/src/error_handle.cpp
+++ b/internal/c/libqb/src/error_handle.cpp
@@ -292,6 +292,73 @@ static void exit_after_fatal_error() {
     libqb_exit(0);
 }
 
+// QB64PE_NOPROMPT suppresses the dialogs that report a runtime error, which
+// avoids the need to interactively interact with the program on unhandled
+// errors.
+//
+//   =y              Reports to stderr and does not prompt, then exits.
+//
+//   =Continue       Reports to stderr and does not prompt, then for
+//                   non-critical errors we will continue to the next line.
+//
+//   unset or other  Reports and prompts as usual
+//
+// A critical error ends the program under either y or continue.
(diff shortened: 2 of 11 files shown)


RhoSigma-QB64 (approved):
Approved, but you still need to document the new behavior and `.noprompt` extension in `docs/testing.md`.

mkilgore:
> Approved, but you still need to document the new behavior and `.noprompt` extension in `docs/testing.md`.

Updated :)

a740g (approved):

QB64Petr (approved):

CookieSnippets:
For what it is worth:

Been lurking (and being amazed at the work being done!!!) in the very distant background, but this one _irked my urk_ a bit as it is one big _pet peeve_ of mine (both learned the hard way, from literature, and from far more experienced oldies who told me):

The naming of those _substates_ (or whatever they are called) is extremely confusing and very ambiguous, and I think it will come back to bite somebody in the behind in the future. Or at least be confused about it.

> This give a way to tell the runtime that unhandled errors should not bring up an interactive prompt. The output text appears the same as with the prompt, but we do not wait for input and instead use the indicated action:
> QB64PE_NOPROMPT=y
> QB64PE_NOPROMPT=Continue
> 'y' exits the program on an unhandled error,
> 'continue' lets the program continue past it to the next line.

'_y_' and '_Continue_' are not very consistent with each other. Better would be either '_y_' and '_c_', or '_yes_' and '_continue_', to keep it consistent. Thus either use full words, or either use just tokens/letters. Don't mix them up, that is bad convention.

But more importantly: the combination of 'noprompt' with 'y(es)' meaning _"exit on unhandled error"_ makes not much sense.
This is in the same line of ambiguity as those infamous “_Do you want to copy or move these files? -> Yes/No_" memes from the good old days of very bad designed message boxes.

Also just the 'NOPROMPT' on its own says not much. No prompt when ? No prompt for what ?

So, I kindly and with upmost respect, propose a change in names so not to confuse future development. Some suggestions could maybe be:

QB64PE_NOPROMPT_ERROR     (the default one?)
QB64PE_NOPROMPT_ERROR=continue
QB64PE_NOPROMPT_ERROR=exit
or QB64PE_NOPROMPT_ERROR=quit
or QB64PE_NOPROMPT_ERROR=halt
     ('exit' and 'quit' indicate some sort of _graceful_ ending, while 'halt' denotes a much more immediate or aggressive ending)

QB64PE_ONERROR=prompt     (the default one?)
QB64PE_ONERROR=exit   or   quit   or   halt
QB64PE_ONERROR=continue

anyways...
If this is in the wrong place, or even out of line: in that case, my apologies, and please disregard....

mkilgore:
@CookieSnippets Yes I agree with pretty much all of that, ultimately I don't really intend for this to be documented externally so I was less concerned about the specific API. I added it as a way to allow the test infrastructure to test the error handling without hanging, as I'm planning to introduce a couple of things soon that need tests around that.

That said if you actually see this as something you'd use then I could go back some-point soon and improve it.

CookieSnippets:
@mkilgore
No worries if this is just meant for internal testing stuff. As long as you still know what's what in a few months/years from now I suppose.... I assume there are MUCH bigger fish to fry and being fried at the moment than this...... Also, I admit I didn't do ANY research regarding this, I just saw it zipping by in my mail (I have notifications turned on for pretty much everything) and it poked that irk in me ;-)

PR #765, opened 2026-08-25
Keyboard & mouse input fixes

### LINKED ISSUE #283: Fix _KEYHIT and _DEVICES to properly work with all keyboard keys on Windows
The basic idea here is to replace the keyboard events we get from GLUT (which just don't work that well, some keys don't send various events) with directly handling the keyboard events from Windows. That should allow us to get accurate reports on keydown and keyup events for `_KEYHIT` and `_DEVICES`.

See the screenshots from Discord:
image
image
This one is in reference to `_DEVICES` behavior:
image

### LINKED ISSUE #418: _BUTTON and _DEVICEINPUT does not register keyboard inputs on Linux
_BUTTON and _DEVICEINPUT command do not register inputs on the keyboard in Linux. 

**To Reproduce**
Using code from the help.

```basic
FOR i = 1 TO _DEVICES
    PRINT STR$(i) + ") " + _DEVICE$(i) + " Buttons:"; _LASTBUTTON(i); ",Axis:"; _LASTAXIS(i); ",Wheel:"; _LASTWHEEL(i)
NEXT

DO
    d& = _DEVICEINPUT
    IF d& THEN '             the device number cannot be zero!
        PRINT "Found"; d&;
        FOR b = 1 TO _LASTBUTTON(d&)
            PRINT _BUTTONCHANGE(b); _BUTTON(b);
        NEXT
        FOR a = 1 TO _LASTAXIS(d&)
            PRINT _AXIS(a);
        NEXT
        FOR w = 1 TO _LASTWHEEL(d&)
            PRINT _WHEEL(w);
        NEXT
        PRINT
    END IF
LOOP UNTIL INKEY$ = CHR$(27) 'escape key exit

END
```

**Expected behavior**
Input from the mouse and keyboard should register. Mouse is the only input registered.

**Additional context**
Compiled and run on QB64PE v3.10.0 64bit. 
Has been tested on Linux Mint and Ubuntu and both have the same result.

### LINKED ISSUE #514: Mac: Command-Tabbing away from editor breaks touchpad and mousewheel scrolling
**Describe the bug**
I'm on Sonoma 14.4.1 and QB64 PE 3.13.1. I found https://github.com/QB64-Phoenix-Edition/QB64pe/pull/468 which fixed what I imagine is the behavior of the scroll wheel inside any program created with QB64PE, but it appears this fix has not been implemented inside the editor as you cannot scroll code or help files without using keyboard keys or dragging the scrollbar.

**To Reproduce**
1. Load QB64 PE on MacOS and enter a command or keyword.
2. Select the keyword and press F1 to open the help screen.
3. Try using the mousewheel and touchpad to scroll. No effect.
4. Go back to the editor and open a lengthy file of code.
5. Try to scroll up and down in the editor with just the scroll wheel. Try the touchpad. Nothing!

**Expected behavior**
It should scroll down a few lines at a time, like dragging the scrollbar with your cursor.

a740g:
This PR addresses several bugs, both new and old. Some were introduced with the recent GLFW swap, while others had been there for a while.

Dead key fixes by @QB64Petr 
malloc()/free() fixes by @grymmjack 

Since automated tests for mouse and keyboard input aren't possible without human interaction, I've included a list of test programs I used to reproduce and catch many of these issues.

[input_tests.zip](https://github.com/user-attachments/files/31403090/input_tests.zip)

Closes #514
Closes #418
Closes #283


QB64Petr:
I tested the current main locally merged with PR #765 on Windows with a Czech keyboard layout.

The low-level AltGr input itself is working correctly after the GLFW port. _KEYHIT, INKEY$, INPUT and LINE INPUT all receive the expected AltGr characters correctly. For example, AltGr+V produces @, AltGr+F produces [, etc.

The remaining problem is in the IDE. The IDE still assumes the old behavior where AltGr appears as Ctrl+Alt. With GLFW, AltGr is exposed as Right Alt, so the IDE treats AltGr as a normal Alt key and either discards the character or activates menus.

I found three places in:

source/ide/ide_methods.bas

that need to exclude _KEY_RALT.

At line 1579, change:

IF NOT idealthighlight AND KALTPRESS = -1 AND NOT KCTRL THEN

to:

IF NOT idealthighlight AND KALTPRESS = -1 AND NOT KCTRL AND NOT _KEYDOWN(_KEY_RALT) THEN

This prevents pressing AltGr by itself from activating/highlighting the IDE menu.

At line 1932, change:

IF KALT AND UCASE$(K$) = a$ THEN

to:

IF KALT AND NOT _KEYDOWN(_KEY_RALT) AND UCASE$(K$) = a$ THEN

This prevents AltGr combinations from opening menus. For example, before this change AltGr+R opened the Run menu and AltGr+H opened Help.

At line 4379, change:

IF KALT AND NOT KCTRL AND NOT AltSpecial THEN GOTO specialchar

to:

IF KALT AND NOT KCTRL AND NOT AltSpecial AND NOT _KEYDOWN(_KEY_RALT) THEN GOTO specialchar

This prevents valid AltGr characters from being discarded by the main IDE editor.

After applying these three changes I tested:

AltGr character input in the IDE         OK
AltGr alone does not activate menu  OK
AltGr+R does not open Run           OK
AltGr+H does not open Help          OK
normal Alt menu behavior            OK
_KEYHIT with AltGr                  OK
INKEY$ with AltGr                   OK
INPUT with AltGr                    OK
LINE INPUT with AltGr               OK

So the GLFW keyboard backend appears to be generating the AltGr characters correctly. The regression is caused by the IDE still treating Right Alt / AltGr as ordinary Alt in these three locations.

All three changes are only in:

source/ide/ide_methods.bas

mkilgore:
@a740g, @QB64Petr - It's probably easier to just exclude RALT as being treated as a real ALT key here: https://github.com/QB64-Phoenix-Edition/QB64pe/blob/3877e5ce3106191be1ea553456e967a4e998b9ee/source/ide/ide_methods.bas#L18883

Then `KALT` only means left-alt and we don't need to add extra checks everywhere. We could probably also remove the `AND NOT KCTRL` that's everywhere since those are just checks for `ALT+CTRL` meaning `AltGr` on Windows, which doesn't currently happen anymore.

This also fixes things on Linux (and maybe Mac OS), it was previously broken there since it didn't report the `ALT+CTRL` combo.

FWIW we also already have some tests that do keyboard emulation on Windows: https://github.com/QB64-Phoenix-Edition/QB64pe/blob/main/tests/compile_tests/keyboard/sendinput.bm

Potentially it could be used for testing some of this, I used it for testing _DEVICES with keyboard in the past.

a740g:
That does seem easier @mkilgore

a740g:
@QB64Petr I get what you are saying. However, I think accurately detecting if a user has a non-US layout keyboard is not easy. Guessing this based on the OS language is not correct IMO. I mean a user could set the OS to Spanish but use a physical US keyboard for example.

I think a better idea would be to add a Keyboard layout toggle in `Option > Language` maybe that allows the user to decide. If they choose "US Keyboard", then we can check both left and right alt, otherwise just the left alt.

What do you think?

QB64Petr:
@a740g  I'll try to write a function for this so that the IDE gets information about whether the keyboard has a right Alt or AltGr. For now, I'd leave it as it is.

RhoSigma-QB64 (approved):

PR #766, opened 2026-08-30
comma array parameter repair

QB64Petr:
repair (or rather return) to the original state.

This doesn't work now, but after the repair it will work again:

Sub bar (baz( ,) As Long)
    For y = LBound(baz, 1) To UBound(baz, 1)
        For x = LBound(baz, 2) To UBound(baz, 2)
            Print baz(x, y),
        Next
        Print
    Next
End Sub

Sub foo (baz( ,) As Long)
    For y = LBound(baz, 1) To UBound(baz, 1)
        For x = LBound(baz, 2) To UBound(baz, 2)
            baz(x, y) = x + y
        Next
    Next
End Sub

Dim a(3, 3) As Long
foo a()
bar a()


a740g (approved):

CookieSnippets:
Wow, was that ever a possibility? What was/is this good for? That seems so completely wrong. Doesn't that teach the wrong thing? That looks like it should give a syntax error.
Yes -> Sub bar (baz() As Long)
No -> Sub bar (baz( , ) As Long)

CookieSnippets:
Wow I didn't know that was even possible!
What is the purpose of this? I mean, where does that come from?
That seems so wrong (the syntax I mean, not the requirement of dimensions).

In no Q(uick)Basic version was that ever a valid syntax as far as I know. And I looked in old documentation, official Q&A and KBs to make sure, but can't find anything. So, out of interest, can anyone point to where this comes from? Or was this something invented for QB64 while forgetting about the historical existing Baz( X ) syntax?

So, the dimensional requirement thing is ok for sure, and very cool to have still/back * (!!). But the historical syntax was simply Baz(3) for a 3-dimensional array.

**This syntax with a comma feels so very very wrong.
And it is even inconsistent with itself!!
Because if you wanna specify a one-dimensional requirement you would logically use (), which is in conflict with the standard/default unspecified dimension. So, this comma syntax doesn't even have consistent mapping to dimensionality! It is non-injective, and makes for very ambiguous coding habits and bad practice! Because now, what does it mean when you see Baz() being used as a parameter? Was that meant to be 'unspecified', or was that meant to be 'one-dimension requirement'?**

If you MUST keep that syntax for compatibility reasons (though I wonder which ones), then please entertain the idea of an automatic conversion in the IDE to the historically correct and unambiguous Baz(X) syntax. This comma syntax should not be allowed. IMHO.


* from the QuickBasic 3.0 and lower days. In 4.x it was dropped in "favor" for the empty syntax ( ). [Q47124](https://www.betaarchive.com/wiki/index.php?title=Microsoft_KB_Archive%2F47124)

PR #767, opened 2026-08-31
Add 32-bit cross-compilation support

mkilgore:
This adds the ability to use `-f:TargetBits=[32|64]` to change the bitness that QB64-PE will produce a C++ program for. Paired with a cross-platform toolchain (Windows already has this) this can allow a 64-bit version of QB64-PE to compile programs for a 32-bit host. Technically a 32-bit version could also produce 64-bit code, but that has less utility given you can't easily compile or run it.

This is not currently exposed in the IDE, only the command line. The main use-case for this is compiling QB64-PE itself - currently we rely on `./internal/source` being valid to compile on both 32-bit and 64-bit. This comes with a lot of caveats and is likely to get accidentally broken in the future (if it's not already broken in some fashion) as more bitness-dependent features like `As String`, `_Dynamic`, and others are introduced into the QB64-PE code itself.

The larger idea is that once this is available we can start treating `./internal/source` as 64-bit only, 32-bit QB64-PE versions will be made via cross-compilation using the 64-bit bootstrap compiler. We can also clean-up some of the questionable bits used to make the existing 64-bit/32-bit compilation work (Ex. `qbs *`s being stored in a uint64_t on both platforms), which will simplify things for later changes/additions.


QB64Petr (approved):

a740g (approved):

RhoSigma-QB64 (approved):

SteveMcNeill:
I'd love to also see this implemented in the IDE as just a simple set of toggles for folks with 64-bit versions.

[ ] Compile EXE to 64-bit
[ ] Compile EXT to 32-bit

If you click them both, it makes you both a 32-bit and a 64-bit version with "Untitledx86" and "Untitledx64" as names, depending on program name like now.

mkilgore:
> I'd love to also see this implemented in the IDE as just a simple set of toggles for folks with 64-bit versions.
> 
> [ ] Compile EXE to 64-bit [ ] Compile EXT to 32-bit
> 
> If you click them both, it makes you both a 32-bit and a 64-bit version with "Untitledx86" and "Untitledx64" as names, depending on program name like now.

Currently switching between the two requires wiping all the precompiled stuff so building both would be painfully slow. Addressing that would not be fun because it impacts all the third-party libraries we build 🤷‍♂️

a740g:
I agree with Matt. A simple checkbox (in compiler options) when running the 64-bit version of the IDE `[ ] Compile 32-bit executable` should suffice in the future.

CookieSnippets:
> > I'd love to also see this implemented in the IDE as just a simple set of toggles for folks with 64-bit versions.
> > [ ] Compile EXE to 64-bit [ ] Compile EXT to 32-bit
> > If you click them both, it makes you both a 32-bit and a 64-bit version with "Untitledx86" and "Untitledx64" as names, depending on program name like now.
> 
> Currently switching between the two requires wiping all the precompiled stuff so building both would be painfully slow. Addressing that would not be fun because it impacts all the third-party libraries we build 🤷‍♂️

Yes, of course it is going to be slow(er), twice as much in fact. I think that speaks for itself. And anybody who uses it, will probably know that too.

Thus, why would that matter? If one wants to compile for both, they would ALREADY be doing everything twice, but manually. Which is far more prone to errors, and quite frankly annoying if you need to do it a few times. Having BOTH options would be a big quality of life addition in those cases, and SAVING time instead.... and REDUCING ERRORS/mishaps.

Of course, this is not an issue of the upmost importance either way.
But still, dismissing a double option like that, just because _"it is slow"_ seems questionable.
(speaking from a user POV of course.... not from the POV of the wizards whom are tasked to implement this)

mkilgore:
> Yes, of course it is going to be slow(er), twice as much in fact. I think that speaks for itself. And anybody who uses it, will probably know that too.
> ...
> Of course, this is not an issue of the upmost importance either way. But still, dismissing a double option like that, just because _"it is slow"_ seems questionable.

The issue is the "wiping all the precompiled stuff" part I mentioned :P It would not just be twice as slow but rather much slower than that.

When you build a program with QB64-PE, all the stuff making up `libqb` is built once and kept around, that allows further compilations to be faster as we simply link the precompiled `libqb` into your program. Swapping bitness requires wiping all of that precompiled stuff so now you're stuck building it twice on every compilation, rather than only once rarely.

CookieSnippets:
> > Yes, of course it is going to be slow(er), twice as much in fact. I think that speaks for itself. And anybody who uses it, will probably know that too.
> > ...
> > Of course, this is not an issue of the upmost importance either way. But still, dismissing a double option like that, just because _"it is slow"_ seems questionable.
> 
> The issue is the "wiping all the precompiled stuff" part I mentioned :P It would not just be twice as slow but rather much slower than that.
> 
> When you build a program with QB64-PE, all the stuff making up `libqb` is built once and kept around, that allows further compilations to be faster as we simply link the precompiled `libqb` into your program. Swapping bitness requires wiping all of that precompiled stuff so now you're stuck building it twice on every compilation, rather than only once rarely.

Yeah, I know... but still... maybe in the (far?) future you could make provisions for separating the needed \internal\whatever paths, one for 64bit another for 32bit, so you don't need to remove everything every time. 😉 😉 
Anyways, disregard if my ignorance of the (much much more complicated than how I suggest it) technicalities makes little sense though... I wonder how many would compile in both all the time anyways, but it would be a nice QOL option to have.... some day.... and then maybe 16-bit real-mode support for the sake of it, so we can write true DOS stuff for our old bricks ... (Im already getting my XT warmed up and ready again)... ROFLOL 😉

SteveMcNeill:
write a simple flag to disk:  LastCompiled = 64

Then when someone wants to compile with the last compiled version, it doesn't have to purge anything as long as it matches.
If I last compiled a 64-bit EXE, the c++ libs are already in 64-bit format and good to go.  No need for those to recompile.

If both options are set for 32 and 64, then compile in whatever was last used first (such as 64-bit as above, so no recompilation necessary), then compile in 32-bit and set the LastCompiled = 32.

Only when the user swaps what they want to do, would they have to do that recompile for the c-libraries, and they'd never need to do it more than once.  When possible, you'd use the existing c binaries without worry.  If not, then you're going to have to recompile those anyway just by manually changing the toggle.   

At worst, you do the recompile *once* to produce both EXEs.  At best, you have one option checked, keep it checked, and then never do a recompile as that LastCompiled will always match what you have set.

Seems to me like it'd just simplify the process and take any/all headache out from the end user experience and not add noticeable overhead in the end, unless someone just wanted to make both versions at the same time.

mkilgore:
@CookieSnippets - Yes that's the eventual goal, really the ideal state is that we ship all of `libqb` precompiled, then the whole problem goes away (along with a bunch of other problems). The only real reason we don't already is because `libqb.cpp` has too many `#ifdef` flags to reasonably provide every combination. That being the case though, it's too complicated to keep the compiled 32-bit and 64-bit intermediate files separate because the third-party libraries we use have their own build systems that don't support it.

@SteveMcNeill - That's what the code already does, see the `.qb64_target_bits` file. Ultimately you're still stuck recompiling everything at least once if you were to have both enabled.

Issue #23, opened 2022-05-03
`X ^ -NOT Y` is considered invalid syntax


mkilgore:
`X ^ -NOT Y` is valid syntax in QB45, but gives a syntax error in QB64 due to it not understanding the `NOT` operator can come after `-` within exponentiation (It thinks there has to be a number). It does correctly understand `X ^ NOT Y` and `X ^ - Y`, so it's just the combination negating the NOT that confuses the parser. The opposite `X ^ NOT -Y` is seemly parsed correctly.

RhoSigma-QB64:
It works as `X ^ (-NOT Y)` which is at least compliant to the [Wiki](https://qb64phoenix.com/qb64wiki/index.php/Mathematical_Operations#Notes) (works that way already in old SDL versions). Just stumbled over this while checking/cleaning/fancy things, those notes are in there for ages, but probably forgotten even by the former developers. They just noticed negative exponents didn't work and tried to fixed it so things like `X ^ -Y` work now without paranthesis. But the fix isn't complete.

QB64Petr:
Bug repaired in #769

Issue #498, opened 2024-06-01
_BIT arrays do not clear all bits on resize


flukiluke:
**Describe the bug**
When performing REDIM _PRESERVE on an array of _BIT, bits in the same byte as a preserved bit are not cleared.

**To Reproduce**
Test case:
```
REDIM a(0 TO 4) AS _BIT * 2
a(0) = -1
a(1) = -1
a(2) = -1
a(3) = -1
a(4) = -1
REDIM _PRESERVE a(0 TO 1) AS _BIT * 2
REDIM _PRESERVE a(0 TO 4) AS _BIT * 2
PRINT a(0); a(1); a(2); a(3); a(4)
```
Indices 2, 3, 4 should be 0 but 2 & 3 are still -1. Note that indices 0 through 3 all share a byte but a(4) is in its own byte, correctly zeroed.

Raven-Singularity:
Just a slight bit of memory corruption.

I've only ever used _BIT for TRUE/FALSE, so haven't run into this issue myself.

PR #769, opened 2026-09-01
Fix compilation issue (UDT arrays) and unary NOT exponent precedence …

QB64Petr:
This PR fix issue #23 and this bug: https://qb64phoenix.com/forum/showthread.php?tid=4888&pid=43459#pid43459


source/utilities/type.bas
```diff
@@ -2112,100 +2112,175 @@ FUNCTION udt_array_member_bytes& (element)
...
+END SUB
+
+SUB clear_udt_with_varstrings (n$, udt, buf, base_offset)
+    ' CLEAR keeps live qbs objects and resets only their lengths, while numeric and fixed-
+    ' length storage is zeroed. Emit the same member-wise behavior through compact runtime
+    ' loops so CLEAR generation remains bounded for large nested inline member arrays.
+    AppendLegacyUDTVarClear n$, udt, buf, LTRIM$(STR$(base_offset)), 0
+END SUB
+
+' Emit legacy inline-layout variable-STRING initialization at an offset expression.
```
mkilgore: Why is it called 'legacy'?

source/utilities/type.bas
```diff
@@ -2112,100 +2112,175 @@ FUNCTION udt_array_member_bytes& (element)
 END FUNCTION
 
 SUB initialise_udt_varstrings (n$, udt, buf, base_offset)
+    ' Keep this established entry point for every legacy inline-layout scalar UDT caller.
```
mkilgore: I don't feel these functions are necessary to keep personally.

RhoSigma-QB64 (approved):

SteveMcNeill (approved):

QB64Petr:
Refactor nested array codegen and fix #498

Refactor nested TYPE array lifecycle and code generation in
qb64pe.bas and type.bas.

- replace compile-time per-element code expansion with compact
  generated runtime loops
- consolidate duplicated ownership, descriptor, reset, and
  REDIM/_RETAIN traversal helpers
- preserve legacy inline, _Static, and _Dynamic array semantics
- harden descriptor/owner copy and self-assignment handling
- fix packed _BIT REDIM _PRESERVE/_RETAIN behavior
- add regression tests for nested arrays, ownership, descriptors,
  codegen, and packed _BIT arrays

Fixes #498

source/qb64pe.bas
```diff
@@ -14670,8 +14634,131 @@ END FUNCTION
...
+'deliberately does not know whether that payload is a qbs owner, a packed _BIT value, a
+'descriptor-aware UDT, or raw fixed-size storage. Keeping payload semantics at the caller
+'prevents ownership rules from being merged merely because the multidimensional walk is common.
+'Compile-time loops below scale only with the number of array dimensions. The number of
+'preserved elements is traversed by the generated C++ for(;;) loop at runtime.
+SUB AppendRetainCoordinateWalk (buf, dimension_count, payload_code$)
+    DIM walk_dim AS LONG
+    DIM desc_slot AS LONG
+
+    WriteBufLine buf, "if (preserve_any){"
```
mkilgore: These need to be `WriteBufLineCpp` so that they don't mess up the debug information.

QB64Petr:
The change to setup_win.cmd fixes a locale-dependent 64-bit Windows detection bug.

The script previously checked:

powershell -NoProfile -Command "if ((Get-WmiObject Win32_OperatingSystem).OSArchitecture -eq '64-bit') { exit 0 } else { exit 1 }"

Win32_OperatingSystem.OSArchitecture returns a localized string. For example, on Czech Windows it returns 64bitový instead of 64-bit. As a result, a real 64-bit Windows system was incorrectly detected as unsupported 32-bit Windows and the setup process aborted.

The check was changed to:

powershell -NoProfile -Command "if ([Environment]::Is64BitOperatingSystem) { exit 0 } else { exit 1 }"

[Environment]::Is64BitOperatingSystem returns a Boolean value and is therefore independent of the Windows display language.

This does not change the intended build policy: the source bootstrap still requires a 64-bit Windows host, while both 64-bit and 32-bit QB64-PE targets can be built. It only makes the host architecture detection reliable on non-English Windows installations.

mkilgore:
> The change to setup_win.cmd fixes a locale-dependent 64-bit Windows detection bug.
> 
> The script previously checked:
> 
> powershell -NoProfile -Command "if ((Get-WmiObject Win32_OperatingSystem).OSArchitecture -eq '64-bit') { exit 0 } else { exit 1 }"
> 
> Win32_OperatingSystem.OSArchitecture returns a localized string. For example, on Czech Windows it returns 64bitový instead of 64-bit. As a result, a real 64-bit Windows system was incorrectly detected as unsupported 32-bit Windows and the setup process aborted.
> 
> The check was changed to:
> 
> powershell -NoProfile -Command "if ([Environment]::Is64BitOperatingSystem) { exit 0 } else { exit 1 }"
> 
> [Environment]::Is64BitOperatingSystem returns a Boolean value and is therefore independent of the Windows display language.
> 
> This does not change the intended build policy: the source bootstrap still requires a 64-bit Windows host, while both 64-bit and 32-bit QB64-PE targets can be built. It only makes the host architecture detection reliable on non-English Windows installations.

Unfortunately I avoided Is64BitOperatingSystem because it's not available on Win7 :-/ We'll probably need to find another solution.

RhoSigma-QB64:
> > The change to setup_win.cmd fixes a locale-dependent 64-bit Windows detection bug.
> > The script previously checked:
> > powershell -NoProfile -Command "if ((Get-WmiObject Win32_OperatingSystem).OSArchitecture -eq '64-bit') { exit 0 } else { exit 1 }"
> > Win32_OperatingSystem.OSArchitecture returns a localized string. For example, on Czech Windows it returns 64bitový instead of 64-bit. As a result, a real 64-bit Windows system was incorrectly detected as unsupported 32-bit Windows and the setup process aborted.
> > The check was changed to:
> > powershell -NoProfile -Command "if ([Environment]::Is64BitOperatingSystem) { exit 0 } else { exit 1 }"
> > [Environment]::Is64BitOperatingSystem returns a Boolean value and is therefore independent of the Windows display language.
> > This does not change the intended build policy: the source bootstrap still requires a 64-bit Windows host, while both 64-bit and 32-bit QB64-PE targets can be built. It only makes the host architecture detection reliable on non-English Windows installations.
> 
> Unfortunately I avoided Is64BitOperatingSystem because it's not available on Win7 :-/ We'll probably need to find another solution.

Most of the PowerShell stuff isn't woking on Win7, it only has PS v2.0, why not simply using the native environment variables PROCESSOR_ARCHITECTURE and PROCESSOR_ARCHITEW6432 for it, simply ask AI how the possible combinations are and how to figure the systems bit-ness from it.

QB64Petr:
Once the tests finish running, I'll either fix the bugs or write a version of setup_win.cmd without PowerShell.

CookieSnippets:
Yeah, don't use PowerShell for this. It is right there in the environment variables:
All these should work on all Windows versions (post XP that is), even when running in 32-bit cmd host under 64bit OS:

```
@ECHO OFF
IF "%PROCESSOR_ARCHITECTURE%"=="AMD64" GOTO 64BIT
IF "%PROCESSOR_ARCHITEW6432%"=="AMD64" GOTO 64BIT
IF "%PROCESSOR_ARCHITECTURE%"=="ARM64" GOTO 64BIT
IF "%PROCESSOR_ARCHITEW6432%"=="ARM64" GOTO 64BIT
IF "%PROCESSOR_ARCHITECTURE%"=="IA64" GOTO 64BIT
IF "%PROCESSOR_ARCHITEW6432%"=="IA64" GOTO 64BIT

:32BIT
ECHO 32-bit OS
SET "OS_BIT=32"
EXIT /B 1
GOTO :EOF

:64BIT
ECHO 64-bit OS
SET "OS_BIT=64"
EXIT /B 0
GOTO :EOF
```
----------------------------------------------------------------------------------------

Or shorter, without the fuzz:

```
@ECHO OFF
REM only returns errorlevel 0 when 64bit, errorlevel 1 when 32bit
IF "%PROCESSOR_ARCHITECTURE%"=="AMD64" EXIT /B 0
IF "%PROCESSOR_ARCHITEW6432%"=="AMD64" EXIT /B 0
IF "%PROCESSOR_ARCHITECTURE%"=="ARM64" EXIT /B 0
IF "%PROCESSOR_ARCHITEW6432%"=="ARM64" EXIT /B 0
IF "%PROCESSOR_ARCHITECTURE%"=="IA64" EXIT /B 0
IF "%PROCESSOR_ARCHITEW6432%"=="IA64" EXIT /B 0
EXIT /B 1
```

----------------------------------------------------------------------------------------

Or shortest I can think of (for Windows 7/Windows Server 2008 R2, and later):

```
@ECHO OFF
IF DEFINED ProgramW6432 EXIT /B 0
EXIT /B 1
```
This one works on the fact that on a 64bit OS, Windows will always create that environment variable. On 32bit, it doesn't (shouldn't) exist.


All adaptable to your needs of course....

RhoSigma-QB64:
> Once the tests finish running, I'll either fix the bugs or write a version of setup_win.cmd without PowerShell.

Oh, and by the way `setup_mingw.bat` has the same problem not running on Windows 7 due to all that PowerShell scripting.
We have a nice programming language at hand with QB64, why not simply utilising it? A simple program returning the output of `_OS$` and we have all we need to determine the architecture and bit-ness. Another small program using the HTTP functionality finding and downloading us the most recent LLVM-MinGW toolchain. These tools could be compiled as 32-bit EXEs, so that they run on both 32/64-bit windows and dropped into `internal`. There they don't hurt and are always available.

RhoSigma-QB64:
@CookieSnippets Your final (shortest) solution will not work as it follows a wrong assumption. I'm on Windows 7 SP 1 64-bit, but I don't see the ....W6432 environment variable here. The other solutions checking for both should work.

CookieSnippets:
> @CookieSnippets Your final (shortest) solution will not work as it follows a wrong assumption. I'm on Windows 7 SP 1 64-bit, but I don't see the ....W6432 environment variable here. The other solutions checking for both should work.

Oh, that is very very strange. So when you do:
```
@ECHO OFF
IF DEFINED ProgramW6432 ECHO ProgramW6432=%ProgramW6432%
IF DEFINED CommonProgramW6432 ECHO CommonProgramW6432=%CommonProgramW6432%
```
Nothing is shown??

It should be there, on the fly created by Windows itself whenever a process starts, like CMD. It doesn't exist as a hard coded environment variable, it is recreated each time.
According to MS itself it should work.... But they didn't specify the service pack though, just _"added since Windows 7 and Windows Server 2008 R2"_ . Maybe it needs SP2 or whatever??

Good to know though!! I'll add a note in my _'list of batch tricks & stuff'_ .

Anyways, yeah, the first ones are the 'traditional'/'old' ways, and works 100% even on Vista. In fact, it should even work on XP, but I can't verify that 100%.

--

FYI: note this is for checking the bitness of the Windows OS itself. If you wanna check the bitness of the running process instead, you need a slightly different logic than just the simple checks shown above.

mkilgore:
> Oh, and by the way `setup_mingw.bat` has the same problem not running on Windows 7 due to all that PowerShell scripting. We have a nice programming language at hand with QB64, why not simply utilising it? A simple program returning the output of `_OS$` and we have all we need to determine the architecture and bit-ness. Another small program using the HTTP functionality finding and downloading us the most recent LLVM-MinGW toolchain. These tools could be compiled as 32-bit EXEs, so that they run on both 32/64-bit windows and dropped into `internal`. There they don't hurt and are always available.

`_OS$` is fixed at compile-time so it can't help here (a 32-bit program running on 64-bit Windows still says 32-bit). For the MinGW download it could work (though it still can't do the unzip), but I mean the general idea is to avoid having to check exe's into the repo, and otherwise you'd have no way to build those programs. Realistically we should be able to write a Win7 compatible version in Powershell that downloads and unzips if someone spent the time to do it.

RhoSigma-QB64:
> > @CookieSnippets Your final (shortest) solution will not work as it follows a wrong assumption. I'm on Windows 7 SP 1 64-bit, but I don't see the ....W6432 environment variable here. The other solutions checking for both should work.
> 
> Oh, that is very very strange. So when you do:
> 
> ```
> @ECHO OFF
> IF DEFINED ProgramW6432 ECHO ProgramW6432=%ProgramW6432%
> IF DEFINED CommonProgramW6432 ECHO CommonProgramW6432=%CommonProgramW6432%
> ```
> 
> Nothing is shown??
> 
> It should be there, on the fly created by Windows itself whenever a process starts, like CMD. It doesn't exist as a hard coded environment variable, it is recreated each time. According to MS itself it should work.... But they didn't specify the service pack though, just _"added since Windows 7 and Windows Server 2008 R2"_ . Maybe it needs SP2 or whatever??
> 
> Good to know though!! I'll add a note in my _'list of batch tricks & stuff'_ .
> 
> Anyways, yeah, the first ones are the 'traditional'/'old' ways, and works 100% even on Vista. In fact, it should even work on XP, but I can't verify that 100%.
> 
> --
> 
> FYI: note this is for checking the bitness of the Windows OS itself. If you wanna check the bitness of the running process instead, you need a slightly different logic than just the simple checks shown above.

Oh yes `ProgramW6432` is there, but not `PROCESSOR_ARCHITEW6432`.

RhoSigma-QB64:
> > Oh, and by the way `setup_mingw.bat` has the same problem not running on Windows 7 due to all that PowerShell scripting. We have a nice programming language at hand with QB64, why not simply utilising it? A simple program returning the output of `_OS$` and we have all we need to determine the architecture and bit-ness. Another small program using the HTTP functionality finding and downloading us the most recent LLVM-MinGW toolchain. These tools could be compiled as 32-bit EXEs, so that they run on both 32/64-bit windows and dropped into `internal`. There they don't hurt and are always available.
> 
> `_OS$` is fixed at compile-time so it can't help here (a 32-bit program running on 64-bit Windows still says 32-bit). For the MinGW download it could work (though it still can't do the unzip), but I mean the general idea is to avoid having to check exe's into the repo, and otherwise you'd have no way to build those programs. Realistically we should be able to write a Win7 compatible version in Powershell that downloads and unzips if someone spent the time to do it.

Indeed, my fault on the `_OS$` thing. For unzipping simply use 7-Zip, as all windows releases are in 7-zip archives, every QB64(PE) user should have it. And even if it's a first time user it's most probably there, as it is a "defacto" standard today.

QB64Petr:
The setup_mingw.cmd compatibility issue can be handled separately in a follow-up PR.

This PR already addresses the setup_win.cmd issue. The current version no longer relies on PowerShell for host architecture detection and instead uses the Windows environment variables intended for this purpose: PROCESSOR_ARCHITECTURE for a native 64-bit command process and PROCESSOR_ARCHITEW6432 when running under WOW64.

At this point, I would like to keep the scope of this PR fixed and avoid adding further unrelated changes.

Please review the current HEAD again and let me know whether you consider the remaining changes acceptable and whether this PR can now be merged.

a740g:
setup_mingw.cmd has to remain a batch file with PowerShell calls or like Matt suggested, we can convert the whole script to PowerShell and ensure it works with PowerShell 2.0 or whatever version shipped with Windows 7.

PR #772, opened 2026-09-05
Store qbs* array elements in pointer-sized slots on each platform

mkilgore:
Previously variable-length STRING arrays were stored in uint64-slots on 64-bit and 32-bit, which was necessary as the generated code in ./internal/source has to compile as both (and offsets into the array are calculated by the QB64-PE compiler and inserted into the C++ code directly).

Now that ./internal/source is only treated as 64-bit this restriction no longer exists and we can simply generate different array offsets for each platform, storing both as arrays of `qbs *` based on the size of `qbs *` for that platform.

The benefit of this is in simplifying the memory-layout - arrays of STRINGs were the only type where the size of their entries in an array were different than the size of that variable declared alone. This is important in preparation for the `mem_layout` stuff I'm working on as it unifies the definition of a STRING inside of arrays vs. outside of arrays, otherwise the difference would have to be part of the layout and exposed to callers.

### DIFF
--- a/source/qb64pe.bas
+++ b/source/qb64pe.bas
@@ -8546,7 +8546,13 @@ DO
                 n$ = RTRIM$(id.callname)
                 bytesperelement$ = _TOSTR$((id.arraytype AND UDTMASK) \ 8)
                 udt = 0
-                IF id.arraytype AND ISSTRING THEN bytesperelement$ = _TOSTR$(id.tsize)
+                IF id.arraytype AND ISSTRING THEN
+                    IF id.arraytype AND ISFIXEDLENGTH THEN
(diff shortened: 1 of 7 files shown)


RhoSigma-QB64 (approved):

QB64Petr (approved):

QB64Petr:
I think the cleanest way forward would be for you to merge your PR first.

Once your changes are in main, I’ll update my work #769 on top of them, make any necessary adjustments, and rerun the relevant tests. After that, I’ll push the updated version of my PR and only then consider it ready for merge.

That should keep things simple and avoid us trying to merge two overlapping versions of qb64pe.bas at the same time. Thanks!

PR #776, opened 2026-09-12
_ArrayCopy statement

QB64Petr:
Add `_ArrayCopy` statement

This adds the `_ArrayCopy` statement for copying elements between arrays.

`_ArrayCopy` supports one-dimensional and multidimensional arrays and provides four copy forms depending on whether the source uses the whole array or a selected range, and whether the destination preserves the source indices or starts at an explicitly specified destination index.

Supported element types include byte-aligned numeric types, fixed-length strings (`STRING * n`), and variable-length `STRING` arrays.

Bounds are validated before writing to the destination. Overlapping copies are handled safely. Variable-length `STRING` arrays use ownership-aware snapshot/deep-copy logic so overlapping copies do not corrupt source strings.

Syntax

_ArrayCopy source() TO destination()

_ArrayCopy source(sourceLow TO sourceHigh) TO destination(destinationStart)

_ArrayCopy source(sourceLow TO sourceHigh) TO destination()

_ArrayCopy source() TO destination(destinationStart)

The same forms also apply to multidimensional arrays, for example:

_ArrayCopy source(x1 TO x2, y1 TO y2) TO destination(dx, dy)

Examples:
1. Whole source array to whole destination array

DIM a(0 TO 50) AS STRING
DIM b(0 TO 50) AS STRING

_ARRAYCOPY a() TO b()

This copies the complete source array while preserving its indices.

Equivalent index mapping:

a(0)  -> b(0)
a(1)  -> b(1)
...
a(50) -> b(50)

For compatible whole arrays, this is similar in purpose to:

b() = a()

2. Source range to an explicit destination start index
DIM a(0 TO 50) AS STRING
DIM b(0 TO 200) AS STRING

_ARRAYCOPY a(20 TO 40) TO b(50)

The selected source range is copied contiguously beginning at the specified destination index.

a(20) -> b(50)
a(21) -> b(51)
a(22) -> b(52)
...
a(40) -> b(70)

The original source indices are not preserved.

3. Source range to the destination array while preserving indices
DIM a(0 TO 50) AS STRING
DIM b(0 TO 40) AS STRING

_ARRAYCOPY a(20 TO 30) TO b()

The selected source range is copied using the same indices in the destination.

a(20) -> b(20)
a(21) -> b(21)
...
a(30) -> b(30)

Only the selected range is copied.

4. Whole source array to an explicit destination start index
DIM a(0 TO 50) AS STRING
DIM b(0 TO 100) AS STRING

_ARRAYCOPY a() TO b(30)

The complete source array is copied contiguously beginning at the specified destination index.

a(0)  -> b(30)
a(1)  -> b(31)
a(2)  -> b(32)
...
a(50) -> b(80)

The original source indices are not preserved.

Quick reference
source()          TO destination()
    Copy the whole array and preserve source indices.

source(lo TO hi)  TO destination(start)
    Copy a selected range beginning at a new destination index.

source(lo TO hi)  TO destination()
    Copy a selected range and preserve the source indices.

source()          TO destination(start)
    Copy the whole source array beginning at a new destination index.

When destination() is used, source indices are preserved.

When destination(start) is used, the copied elements are placed contiguously beginning at start.

For multidimensional arrays, the same rule applies independently to every dimension.


QB64Petr:
The implementation has now been refactored so the heavy _ArrayCopy logic lives in the shared libqb C++ runtime instead of being expanded into the generated C++ code at every call site.

This covers all 1D and multidimensional forms for numeric types, fixed-length strings, and variable-length strings, including overlap-safe and ownership-aware copies.

I also cleaned up the old generated-code helpers, refactored the C++ runtime to remove duplicated validation/mapping logic, and added additional overflow/SIZE_MAX hardening for 32-bit builds.

Final regression tests pass on both 32-bit and 64-bit builds.

internal/c/libqb/src/array-copy.cpp
```diff
@@ -0,0 +1,954 @@
...
+    qbs **destination_slots = (qbs **)destination[0];
+
+    for (uint64_t i = 0; i < count; ++i) {
+        if (!snapshot.make(i)) return;
+        qbs_set(snapshot[i], source_slots[source_offset + i]);
+        if (is_error_pending()) return;
+    }
+
+    for (uint64_t i = 0; i < count; ++i) {
+        qbs_set(destination_slots[destination_offset + i], snapshot[i]);
```
mkilgore: Is there a reason we can't move the `qbs *`s from the 'snapshot' directly into the destination, skipping the need to call `qbs_set()`?

Separately, is the 'snapshot' actually necessary? It seems like it should be possible to determine the correct the copy direction and do a `qbs_set` directly from the source. As it is the snapshot is basically the same size as a third array.

a740g:
@QB64Petr overall this looks good to me.
Can you add some multi-dimension test cases as well?
Also how about UDT arrays and arrays of UDTs?

QB64Petr:
@a740g I’ll add a few more tests; copying from UDT arrays to standard arrays of the same type - and vice versa - is supported, and copying an entire UDT type is now possible in current version using the `=` operator (this was included in a some previous PR).

RhoSigma-QB64 (approved):

PR #777, opened 2026-09-16
Speed up transpiler: hash-index manageVariableList (O(n²) → O(n))

grymmjack:
## Summary

The BASIC→C++ **translate phase** dominates compile time on large programs, and a
big chunk of it was an O(n²) hotspot in `manageVariableList`: it linearly scanned
**every variable created so far** on every call to find an existing entry by
`cname`. Profiling a large program (a 175k-line project) showed ~1.57 **billion**
`qbs_equal` calls coming from that one scan — 26%+ of total transpile time.

This PR adds a hash index over `usedVariableList().cname` so the lookup is O(1)
average instead of O(n).

## Results (on a 175k-line project)

| | transpile + compile wall time |
|---|---|
| before | **270.7s** |
| after | **207.9s / 202.5s** (two runs) |
| **improvement** | **~24% (~65s)** |

Profiling confirms the mechanism: `qbs_equal` calls dropped from **2.54 billion →
971 million** (−62%), and `manageVariableList` fell out of the top-15 self-time
entirely. What remains on top afterwards is diffuse `qbs_set`/`qbs_free`/`qbs_concat`
string churn (inherent to the string-based transpiler), with no remaining
concentrated hotspot.

## Correctness

- **Generated C++ is byte-identical** before/after: diffed all 5,657 generated
  `main*.txt` / `data*.txt` files for the large program → zero differences. This
  change only affects *how* an existing entry is found, never *what* is generated.
- **Full test suite passes** (`tests/run_tests.sh`, 1699 checks). Two
  `tests/compile_tests/glut/screenhide_*` tests show a runtime diff, but the only
  differing line is `_TITLE$` reading back garbage from the window manager on the
  test machine — every other GLFW call matches. Since the generated C++ is
  byte-identical, both compilers produce the identical executable and fail those
  two identically; it's environmental, not from this change.

## Implementation

`source/qb64pe.bas` only (+46 / −3):

- new `usedVarList.cnameHashNext` field chains entries within a bucket
- new `cnameHashHead()` bucket array (65536 buckets), re-REDIM'd per compile
  alongside `usedVariableList` so both reset together
- new `cnameHashBucket&()` — a **full-string djb2** hash. (Note: I deliberately did
  *not* reuse `HashValue&()`, which keys on first/last chars + length. Generated
  cnames share those heavily — `__ARRAY_LONG_...`, `_SUB_..._INTEGER_...` — so it
  collides catastrophically and rebuilds the linear scan. A full-string hash keys
  on the distinguishing middle.)
- the scan in `manageVariableList` becomes an O(1)-average bucket-chain walk; new
  entries prepend to their bucket at insert. On a miss, `i` is set explicitly to
  `totalVariablesCreated + 1` (the value the old `FOR` loop left behind, which the
  add path relies on).

Happy to tune the bucket count or hash if you'd prefer something else.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


grymmjack:
@mkilgore here is the PR i mentioned in the DM - it sounds like you already have a better solution than this though. feel free to reject/close or do whatever with this PR.

mkilgore:
@grymmjack - #778 includes this along with other improvements.

Though I didn't see your note about the hash function until now. I'm currently keying on first and last 12 long with the length, interestingly I adjusted my hash function and don't see any performance change from hashing the whole string, so 🤷‍♂️ Presumably the buckets are short enough that it doesn't matter. Regardless, I'll fix that.

grymmjack:
> @grymmjack - #778 includes this along with other improvements.
> 
> Though I didn't see your note about the hash function until now. I'm currently keying on first and last 12 long with the length, interestingly I adjusted my hash function and don't see any performance change from hashing the whole string, so 🤷‍♂️ Presumably the buckets are short enough that it doesn't matter. Regardless, I'll fix that.

Thanks @mkilgore really glad for this speed up!

grymmjack:
This is included in #778

PR #778, opened 2026-09-18
Apply various compiler optimizations to speed-up compilation time.

mkilgore:
The below table shows the before/after for `-z` (C++ generation only) of QB64-PE itself and [DRAW](https://github.com/grymmjack/DRAW/tree/main) (another large QB64-PE program) on my machine:

| Program | Before | After |
| :--: | --: | --: |
| QB64-PE | 41s | 8.9s |
| DRAW | 5m 59s | 43s |

For that 'After' DRAW time, it's possible to reduce it further with a minor SUB/FUNCTION reordering to avoid an array argument recompile, then it's down to ~22s. Ideally we would avoid those recompiles without having to manually reorder SUB/FUNCTIONs, but fixing that is messy and outside the scope of what I'm trying to do here.

The above does not include the C++ compilation time so the actual compile time is longer. I have some further changes planned to improve the speed on that side of things.

For details on the specific changes I made please reference the individual commit messages as they cover the details. The majority of it is pretty minor, with the simplebuffer changes being the largest part.


RhoSigma-QB64 (approved):
Please reword `utilities/s-buffer/readme.txt` to reflect it's no longer the full fledged version as in my Libraries collection, but rather a special edition reduced/adapted for the usage inside the QB64-PE sources.

a740g (approved):

QB64Petr (approved):

grymmjack:
AMAZING WORK! Thanks @mkilgore !!!

